PR #1, opened 2025-02-24
Add core features to README

OwnerOfJK:
## Goal
Testing signed commit + pr flow.

Added additional comments to Readme for easier work distribution.

PR #4, opened 2025-03-01
Basic compose w/ caddy and dynamic Caddyfile (based on PROD env var)

cubernetes:
Merge #1 first, then this one.
Titles says it all. If you have a .env in the repo root with PROD=1 and the appropriate DOMAIN='...', then a docker compose up will start Caddy in "production" mode where it will substitute the appropriate variables in the Caddyfile, resulting in TLS Certificate generation and so on. Otherwise, it will just be "development" mode, meaning http://localhost and no Certificate generation.


OwnerOfJK (commented):
I am unable to run the docker container. I receive the following error:

```bash
docker compose up -d
validating /home/jkaller/Desktop/HomeOf42/Core_Curriculum_Projects/Rank06/ft_transcendence/compose.yaml: services.caddy-coraza.develop.watch.0.action services.caddy-coraza.develop.watch.0.action must be one of the following: "rebuild", "sync", "sync+restart"
exit status 15
```

I tried it both on the Linux machine and my mac and get the same error. It should be related to the `compose.yaml` file.

After changing the following it works (though that might not be what you intend to do):

```yaml
    develop:
      watch:
        - path: ./caddy-coraza/Caddyfile.template
          action: sync+restart #exec -> restart
          target: /etc/caddy/Caddyfile.template
          exec:
            command: /caddy-wrapper.sh reload
```

darrenkuro (approved):

OwnerOfJK (approved):

PR #5, opened 2025-03-01
Feature/coraza setup

cubernetes:
Merge #4 first, then this one.
Enables the coraza WAF functionality with the OWASP CRS (Core Rule Set), which works surprising well, as seen with the test cases provided in the comments inside the Caddyfile.template.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,7 @@
+DOMAIN=ft-transcendence.app
+SCHEME=https
+
+PROD=0
+
+DEV_HTTP_PORT=8080
+DEV_HTTPS_PORT=8443
--- a/caddy-coraza/Caddyfile.template
+++ b/caddy-coraza/Caddyfile.template
@@ -2,14 +2,25 @@
 	skip_install_trust # caddy is running as non-root, therefore CA trust installation won't work (and isn't needed anyway inside the container)
 	http_port 8080
 	https_port 8443
+	order coraza_waf first
 }
 
 (static) {
+	coraza_waf {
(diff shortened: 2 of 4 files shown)


darrenkuro:
Not really familiar with Coraza but setup seems straightforward and standard enough. HTTP works fine but is HTTPS also supposed to work? It doesn't on mine, but might be issue with docker routing? 

As a side note: merge order is independent of reviews right, say even if a later merge pull passed all the reviews required, it's still fine as long as the it awaits the previous pulls to be merged before it's confirmed to be merged?

cubernetes:
Good questions.

Q: HTTP works fine but is HTTPS also supposed to work?
A: Yes, since Coraza integrates tightly into Caddy, Caddy first does the TLS termination (decrypting the request using the certificates), and then the Coraza module runs the list of rules (policies) against the request. You can test it on the domain: https://ft-transcendence.app/?exec=/bin/bash, it will give a 403 Forbidden.

Q: It doesn't on mine, but might be issue with docker port routing?
A: Correct, for local development, there is no https (intentionally), although you can 100% enable it, but you will get the dreaded "The site you're connectimg to is insecure" error. I will explain this further in our meeting in 1h

Q: As a side note: merge order is independent of reviews [...] merged before it's confirmed to be merged?
A: Technically yes, but I've enabled a GitHub setting that invalidates the review status once a new commit is pushed (since that commit would be unreviewed). I'm not sure how will this interacts with rebasing, so we'll see today.

darrenkuro (approved):

PR #6, opened 2025-03-03
Feature/sqlite

darrenkuro:
Branched off the most recent `feature/client-setup` so it's up to date (work by Timo - see commits for details - merging the branch with the most updated version of main). 

Added the rudimentary schema.sql and the backend docker container build to correctly let frontend fetch some dummy data (server.js is serving only as an illustration of basic functionality). 

On the frontend side, I tried my best to contain the changes on one single file (test.ts which is also only servers as illustration and would be deleted later) but since I am fetching data I had to change a couple functions to async and add some previously inferred types so that it would transpile. Only funky thing is that I didn't want to make too many functions async so had an or in the router return type for now. 

Also added the backend container on compose file so everything builds and runs correctly, and in the leaderboard page the dummy data could be fetched from database by an example api.


cubernetes (commented):
Saw everything already, but went through it once more. Looks good! (Didn't look at the schema and TS suuuuper thoroughly but should be fine :D)

cubernetes (approved):

PR #7, opened 2025-03-03
Feature/api expansion

bejasch:
Branched from Darrens `sqlite` to add some API endpoints inside the new backend/routes folder.

New requests are: 
GET /friends/:userId - Get all friends details for a specific user: id, username, and avatar_url
POST /friends - Send friend request
PUT /friends/accept - Accept friend request
GET /games - Get all games
GET /games/:id - Get specific game details
GET /tournaments - Get all tournaments
GET /tournaments/:id - Get tournament details
GET /users - Fetch all users' info
GET /users/:id - Fetch a specific user's info by ID
POST /register - User registration (new)
POST /login - User login

With this, I added the packages 'bcrypt' (pw hashing) and 'jsonwebtoken' for authentification - which is open to discussion if it should be used like this in the end.

### DIFF
--- a/backend/database/schema.sql
+++ b/backend/database/schema.sql
@@ -14,25 +14,21 @@ CREATE TABLE IF NOT EXISTS users (
 -- Games table
 CREATE TABLE IF NOT EXISTS games (
     id INTEGER PRIMARY KEY AUTOINCREMENT,
-    tournament_id INTEGER,
+    tournament_id INTEGER,  -- can be NULL if not part of a tournament
     player1_id INTEGER NOT NULL,
     player2_id INTEGER NOT NULL,
-    winner_id INTEGER NOT NULL,
+    winner_id INTEGER,  -- can be NULL if the game didnt end yet
     player1_score INTEGER NOT NULL,
     player2_score INTEGER NOT NULL,
-    status TEXT DEFAULT 'finished',
     created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
     finished_at TIMESTAMP,
     FOREIGN KEY (tournament_id) REFERENCES tournaments(id),
     FOREIGN KEY (player1_id) REFERENCES users(id),
     FOREIGN KEY (player2_id) REFERENCES users(id),
     FOREIGN KEY (winner_id) REFERENCES users(id),
-    CHECK (winner_id = player1_id OR winner_id = player2_id),
+    CHECK (winner_id IS NULL OR winner_id = player1_id OR winner_id = player2_id),
     CHECK (player1_id != player2_id),
-    CHECK (player1_score >= 0 AND player2_score >= 0),
-    CHECK (finished_at IS NOT NULL),
-    CHECK (status IN ('finished', 'pending')),
-    CHECK (status = 'finished' OR (finished_at IS NULL AND winner_id IS NULL))
+    CHECK (player1_score >= 0 AND player2_score >= 0)
 );
 
 -- Tournaments table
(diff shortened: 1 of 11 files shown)


backend/database/schema.sql
```diff
@@ -20,14 +20,19 @@ CREATE TABLE IF NOT EXISTS games (
     winner_id INTEGER NOT NULL,
     player1_score INTEGER NOT NULL,
     player2_score INTEGER NOT NULL,
+    status TEXT DEFAULT 'finished',
```
darrenkuro, line 23: We thought about status the game, but ended up opting out because the finished timestamp achieve virtually same function (null = ongoing, not null = finished.) And wasn't really sure if it'd be called/useful... It is a bit redundant but does make it more comprehensive

backend/database/schema.sql
```diff
@@ -20,14 +20,19 @@ CREATE TABLE IF NOT EXISTS games (
...
     finished_at TIMESTAMP,
     FOREIGN KEY (tournament_id) REFERENCES tournaments(id),
     FOREIGN KEY (player1_id) REFERENCES users(id),
     FOREIGN KEY (player2_id) REFERENCES users(id),
     FOREIGN KEY (winner_id) REFERENCES users(id),
     CHECK (winner_id = player1_id OR winner_id = player2_id),
-    CHECK (player1_id != player2_id)
+    CHECK (player1_id != player2_id),
+    CHECK (player1_score >= 0 AND player2_score >= 0),
+    CHECK (finished_at IS NOT NULL),
```
darrenkuro, line 33: finished_at actually can be null if it's ongoing I think? also btw if it's just not null it doesn't require a check it can be on the same line

backend/database/schema.sql
```diff
@@ -20,14 +20,19 @@ CREATE TABLE IF NOT EXISTS games (
     winner_id INTEGER NOT NULL,
```
darrenkuro, line 20: oh just saw a small bug, winner_id can be null, it's checked at the end of this table, winner_at and finished_at have to be both valued or both null

backend/database/schema.sql
```diff
@@ -43,7 +48,7 @@ CREATE TABLE IF NOT EXISTS tournaments (
     CHECK ((winner_id IS NULL AND finished_at IS NULL) OR (winner_id IS NOT NULL AND finished_at IS NOT NULL))
 );
 
--- Friends table
+-- Friends table: user1_id < user2_id ensures searching for friends is easier
```
darrenkuro, line 51: I was actually thinking to ensure there's no duplicate, such as there would be two entries (A, B) and (B, A); also maybe comment in line for consistency? idk we haven't discussed style yet but just a personal preference

backend/routes/friendRoutes.js
```diff
@@ -0,0 +1,109 @@
...
+export default async function friendRoutes(fastify, options) {
+    const { db } = options;
+  
+    // Get all friends details for a specific user: id, username, and avatar_url
+    fastify.get('/friends/:userId', async (request, reply) => {
+        const { userId } = request.params;
+        try {
+            // Fetch friends where userId is either the smaller (user1_id) or larger (user2_id)
+            const friends = await db.all(
+                `SELECT
```
darrenkuro, line 17: I was thinking maybe using an ORM like prisma or knex or something since raw query doesn't look great. I can refactor it after if you agree.

darrenkuro (commented):
The routes look ok, didn't look too deep into logic yet. Made some inline comments in location but I'm actually not sure how GitHub works, can probably merge before "resolving"? Most are just some relevant thoughts. 

Also as mentioned on Discord, I do think the backend should also be on ts for various reasons. Again, everyone's thoughts welcome. If we agree then maybe we can go ahead to merge this branch and I can take a deep look into the logics while I refactor the routes after, should be pretty simple to do that still.

darrenkuro (approved):

cubernetes:
Agree on the TS refactor :D, the code base will grow, so TS will guard off a lot of technical debt in the future

backend/routes/friendRoutes.js
bejasch: Never used it, but yeah - why not :+1:

backend/database/schema.sql
bejasch: I think, having duplicates for one relation is not a good idea in general. There would be a higher cost to change sth.
I tried to care for this in the API.

backend/database/schema.sql
bejasch: True!

backend/database/schema.sql
bejasch: Didn't double check this section, which should be changed!

backend/database/schema.sql
darrenkuro: Yeah so user1 < user2 will ensure each two user pair can only have one relation on this table, that was the intention

backend/database/schema.sql
darrenkuro: Cool, then we can merge this PR and I will refactor it as I fix these small things, sounds good?

PR #9, opened 2025-03-04
Caddy data dir, makefile, and some cleanup

cubernetes:
# All Features (see commit messages)
1. Add simple Makefile with `prod`, `dev`, and `clean` phony targets
2. Add volume to Caddy container to persist Caddy's data directory which contains the TLS certificates (otherwise, it has to be requested from ZeroSSL everytime and we will get blocked by then)
3. README updates regarding running the compose project
4. Path normalizations
5. Change from Let's Encrypt to ZeroSSL
6. Add BACKEND_PORT to .env.example
7. Change ft_trans.. to ft-trans.. and ftt_... to ft-transcendence-...
8. Add docker compose down to make clean target
9. Rename COPYING -> LICENSE


darrenkuro (approved):
I still don't like the ftt-.. naming is so hard, I'll think of something later maybe. Is it better to just have the full word? 

I was writing a Makefile earlier too only because I feel like it was required in the manual but I looked into it and didn't find it. Always good to have one, but was kind of curious whether it's a requirement for this project (could've sworn it was at one point...)

Also, I think we discussed this yesterday but upon further research I am a little confused again. Is there a particular reason why caddy and frontend are on the same docker container (any reason they have to be)? The standard practice (from what I can find online, chatGPT, other people's repos such as Martin's for instance) is to separate them as far as I can tell, you’ll just have volumes in docker compose, then reverse_proxy frontend also; probably cleaner? Doesn't seem like a big deal either way though, just a thought before I forget - was looking into it because was trying to figure out the acceptable build process on both cleanness and complexity for backend, was kind of a pain :/

cubernetes:
Issue: I still don't like the ftt-.. naming is so hard
Action: I pushed another commit ;), now it's uniform

Issue: I was writing a Makefile earlier too [...] (could've sworn it was at one point...)
Answer: Yeah I thought as well, but maybe only for the Inception project (https://timo.one/inception)?

Issue: Is there a particular reason why caddy and frontend are on the same docker container
Answer: Yeah we discussed this, it was weird. The thing is, if you separate them, you have to create 2 containers instead of 1. That in itself is not bad, but what is odd is that one of the containers is basically just a "builder" container (it exists immediately once it has bundled the project). If that is not a weird thing at all, then YES, let's split it up into 2 containers ASAP and mount the dist folder on both before they become too coupled

darrenkuro:
> The thing is, if you separate them, you have to create 2 containers instead of 1. That in itself is not bad, but what is odd is that one of the containers is basically just a "builder" container (it exists immediately once it has bundled the project, it's not a service). If that is not a weird thing at all, then YES, let's split it up into 2 containers ASAP and mount the dist folder on both before they become too coupled

From what I checked with chatGPT earlier, that was supposed to be the case, frontend container is only serving static files. Couldn't find the conversation earlier so tried to recreate it, let's see if this works: https://chatgpt.com/share/67c73f02-6ea0-8006-875d-978d95f5d3ef

darrenkuro:
Oh also completely irrelevant but since this is the only open PR currently I'll just leave it here: COPYING (in root) should be LICENSE, right? Maybe just change it here since it's so trivial.

darrenkuro:
Oh that's interesting, never really heard of that before (but now that I'm thinking about it I might have seen it occasionally..) Apparently COPYING is most appropriate for GPL which is apparently a type of specific license, and yeah popular for GNU projects/linux distro. It feels a little out of place here, I thought you forgot to rename it lol Gerunds just don't feel like proper nouns to me haha unless you are used to it probably; sure I guess if you don't mind I will just change it.

cubernetes:
Actually let me do it, it actually fits quite nicely into this PR :D, don't need to clutter another PR with this file renaming

bejasch (approved):

PR #14, opened 2025-03-07
Feature/database api

darrenkuro:
- Switch to migration with drizzle-kit instead of checking and dealing with sqlite file directly, and integrated into docker images and containers correctly. 
- Removed salt as a field, apparently the library writes salt into hash so no need for a separate field.
- Restructured some backend functions and classes (mostly App, database, and services) for cleaner codes. Still in the process of "trial 'n error"-ing, haven't landed on any that's super satisfying.
- Added authentication as its own service. 
- Fixed the missing /public on web.


darrenkuro:
- This should be ready to be merged after review. 

Some more minor changes:

- WebSocket basic structure and integrated correctly into the Fastify instance now. 
- Configured some logging for Fastify, now instead of console.log, you'd always just get a logger from the instance, depending on where you are on the backend architecture, this should be very available. 
- Database migration and seeding now should be completely correct (hopefully, this process is more prone to bugs than I realized.)
- Add a couple simple targets in Makefile, notably `nodemon` for hot reloading that we discussed yesterday. Currently only on the backend, not the perfect solution but seems like it works (/dist is mounted and I am using another process to re-bundle outside of Docker, while inside the docker nodemon is watching the /dist/index.js so anything change in src triggers outside bundler -> new /dist/index.js -> trigger nodemon in Docker to restart.) I am pretty certain there should be much more elegant solution. Also, as I was struggling through this process I am starting to really appreciate why separation of concern is DEFINITELY good practice (and should be closely followed as much as possible). For instance, I was trying to make frontend work with hot-load as well but it feels much more difficult somehow (probably because I'm getting tired and more stupid as the sun comes up....). but also, if frontend and caddy are two separate containers, which seemed to be a purposeless overkill at the time, would have made this process at least 10 times easier for me. But as things stand now the web container is a little too difficult for me to make it work, maybe someone more experienced with Docker could try that.
 
Anyways, if no major issues are found please review it and merge soon. :)

darrenkuro:
- The hot reload trick I was attempting might be too hacky and therefore very buggy for different environments, so took out for now and gonna look for better solutions.

bejasch (approved):
Maybe add the 'cp .env.example .env' to the Makefile.

cubernetes:
Ok awesome :), haven't reviewed yet, but will do after this comment. From your points @darrenkuro:
- I'm gonna make a branch where I'll separate the caddy and frontend containers using volumes
- Make a branch to add hooks, including your formatting hook

darrenkuro:
> Ok awesome :), haven't reviewed yet, but will do after this comment. From your points @darrenkuro:
> 
> * I'm gonna make a branch where I'll separate the caddy and frontend containers using volumes
> * Make a branch to add hooks, including your formatting hook

Awesome, while you separate the containers, could you also give dev/watch (hot reload) a try. It'd actually be a very useful feature to have during development and I've been a little stuck to find a clean way to make it working well (even for just the backend...)

cubernetes:
Ok I'll try in a third branch :D

cubernetes:
Also, I'm writing a lot of remarks/comments in the review, can you take a look with 👍🏼 on them if you are gonna make a commit fixing it, or a 🚧 if you don't want to do it and I should do it in another commit on another branch? Or give a comment if it's more complicated

darrenkuro:
> Also, I'm writing a lot of remarks/comments in the review, can you take a look with 👍🏼 on them if you are gonna make a commit fixing it, or a 🚧 if you don't want to do it and I should do it in another commit on another branch? Or give a comment if it's more complicated

Sounds good, I don't see it yet? Are you not doing that inline?

cubernetes:
Whut? I added like 10 comments already haha, don't you see them? They should be visible in this conversation as well if you scroll up

Makefile
```diff
@@ -1,17 +1,72 @@
...
+
+.PHONY: install
+install:
+	npm --prefix=web install
+	npm --prefix=backend install
+
+EXTS := ts json
+SPACE := $(empty) $(empty)
+EXTS_PATTERN := $(subst $(SPACE),|,$(EXTS))  # Convert "ts json" → "ts|json"
+PRETTIER := npx --prefix=backend prettier --write
```
cubernetes, line 64: I really tried to figure out what --prefix does by playing around, and it really seems like you can just remove it, prettier will work from the current working directory from where it is ran anyways. If you *really* want to only do the stuff in the backend folder, you have to `cd` into it I think. At least that's what I could observe from testing it

cubernetes: I know this is ironic because I remember telling you that you can put --prefix haha

README.md
```diff
@@ -31,6 +31,12 @@
 - Log management and observability (ELK + Grafana) - Sonia
 - Game statistics also on Blockchain - John
 
+
+## Debug
+- Websockets: Use wscat to connect to frontend (or backend via container IP):
```
cubernetes, line 36: Nit: Add comment that `backend via container IP` will only work if the backend exposes a port, which it doesn't by default

backend/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:18-alpine AS builder
+FROM node:20-alpine AS builder
```
cubernetes, line 1: I would probably pin (and upgrade) the version even further, so:
FROM node:[23.9.0-alpine3.21](https://hub.docker.com/layers/library/node/23.9.0-alpine3.21/images/sha256-191433e4778ded9405c9fc981f963ad2062a8648b59a9bc97d7194f3d183b2b2) AS builder

backend/Dockerfile
```diff
@@ -7,11 +7,10 @@ RUN npm install
 COPY . .
 RUN npm run build
 
-
 FROM node:20-alpine AS production
```
cubernetes, line 10: Same here:
FROM node:23.9.0-alpine3.21

I would leave out the `AS production`, since this is not a build stage and will also be used during development, not only production

backend/Dockerfile
```diff
@@ -7,11 +7,10 @@ RUN npm install
...
 
 # Install build dependencies for bcrypt and other native modules
-RUN apk add --no-cache --virtual .build-deps \
+RUN apk update && apk add --no-cache --virtual .build-deps \
```
cubernetes, line 13: `apk update` is not needed when you add `--no-cache` to the `apk add` command

backend/src/utils/errors.ts
```diff
@@ -1,54 +1,62 @@
...
+        reply.status(this.statusCode).send(this.payload);
+    }
 }
 
 export class BadRequestError extends CustomError {
-  constructor(message = "Bad request") {
-    super(message, 400);
-  }
+    constructor(message = "Bad request") {
+        super(message, 400, "Bad request");
```
cubernetes, line 28: Why is specifying the http status code message needed? Where will that text be shown?

backend/src/utils/validator.ts
```diff
@@ -0,0 +1,8 @@
+import { BadRequestError } from "./errors";
+
+export const validateId = (input: string): number => {
+    const id = Number(input);
+    if (isNaN(id) || !Number.isInteger(id) || id < 1)
```
cubernetes, line 5: Why is `id = 0` bad? Maybe change to `id < 0`?

Makefile
```diff
@@ -1,17 +1,72 @@
 DC := docker compose
 D := docker
 
+RM := rm -rf
+DB := backend/drizzle/db.sqlite
```
cubernetes, line 5: In compose.yaml, we use `DB_PATH`, but here it is `DB`, maybe normalize and use `DB_PATH` everywhere?

web/Caddyfile
```diff
@@ -0,0 +1,36 @@
+{
+	acme_ca https://acme.zerossl.com/v2/DV90
```
cubernetes, line 2: Will probably change it back to Let's Encrypt by removing these 2 lines, but will do myself

web/src/scripts/components/old_GameSection
cubernetes, line 1: Hm there seems to be no JS extension, is this intentional?

cubernetes (changes requested):
Good in general, mostly nitpicks haha, but also some smaller things I'd like to have changed before

web/src/scripts/components/old_GameSection
darrenkuro: This is either from Ben or John, I take from the file of the name it is the old version that is no longer used. I was thinking we could have an archive or sth that's gitignored and we can all save our draft/old files, but technically they are also all just in git history so....

cubernetes: Yeah then I'd remove the file and keep it in the history. With git commands you can always fuzzy search for the file name or file contents through the history, so all good

Makefile
```diff
@@ -1,17 +1,72 @@
...
+.PHONY: re
+re: clean dev
+
+.PHONY: install
+install:
+	npm --prefix=web install
+	npm --prefix=backend install
+
+EXTS := ts json
+SPACE := $(empty) $(empty)
```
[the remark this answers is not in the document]
cubernetes: Actually not sure if this works, maybe you know better

cubernetes:
Resolved all the conversations that I want to do, also added a 👀 emoji to it to mark that I will do it

backend/src/utils/errors.ts
darrenkuro: Currently it's being send directly in reply object
```
...
        catch (error) {
            return error instanceof CustomError
                ? **_error.send(reply)_**
                : new CustomError(`Failed to fetch game by ID`).send(reply);
        }
```
in constructor one is name one is message (which is usually used) but I was just looking at Error obj and wanted to make it complete. But I will probably scrap these anyway since as I was reading the documentations Fastify has more native ways to handle Errors in similar way so making our own classes here aren't really necessary.

backend/src/utils/validator.ts
darrenkuro: sql (maybe sqlite?) when primary key (id) is set to auto-increment, the starting value is 1, which sort of makes sense.

darrenkuro: But also, again, turns out fastify has more native ways to do data validation. Might try other things. But the takeaway here is that auto increment in sql db starts at 1 (probably doesn't matter too much though, just depends on how you want your api end point to accept what kinds of data and give back what kind of response). Since supposedly these rest apis are internally used and not for clients themselves to use technically, shouldn't matter much either way

backend/src/utils/errors.ts
cubernetes: I see I see, you can mark it as resolved if you want

backend/src/utils/validator.ts
cubernetes: I see, was curious. Can mark as resolved

web/src/scripts/components/old_GameSection
cubernetes: Can mark as resolved

Makefile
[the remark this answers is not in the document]
darrenkuro: Oh I do not (know better), apparently you cannot do this: `SPACE := " "` there's something special about spaces in Makefile. I didn't look too much into it, I was just trying to make it work here as a clutch for formatting, got the cmd from internet/ai and it worked-ish. I full expect formatting process to be changed and handled differently. Not sure in which way though, maybe a pre-commit hook? You'll probably have better ideas for this.

Makefile
```diff
@@ -1,17 +1,72 @@
...
+EXTS_PATTERN := $(subst $(SPACE),|,$(EXTS))  # Convert "ts json" → "ts|json"
+PRETTIER := npx --prefix=backend prettier --write
+
+.PHONY: format format-staged
```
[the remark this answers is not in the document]
darrenkuro: I agree, yeah I wasn't sure if these will end up being used (see the other comment about formatter).

backend/Dockerfile
darrenkuro: Actually only changed this because Docker timed out my account for making too many requests but yeah it makes sense. For both builder and production containers right?

backend/Dockerfile
darrenkuro: Yeah the original is perfectly as it is, but I think when I was trying to do a watch with some sort of hack this was needed for some reason. But it was too messy at the end, I will change it back for now in the next commit and leave the watch mode for future.

Makefile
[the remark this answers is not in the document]
cubernetes: Yeah haha almost guessed, let's keep it like this for now and I'll make pre-commit hooks on another branch

README.md
darrenkuro: ~~Can you add this? I have format on save turned on but I am not sure we want to format .md (though, doesn't hurt?) and it is surprisingly a pain to turn off once I have it on.~~ Never mind I just remember I can just do it in helix lol

Makefile
```diff
@@ -1,17 +1,72 @@
...
+fclean: deepclean
+	$(RM) backend/node_modules/ backend/dist/
+	$(RM) web/node_modules/ web/dist/
+
+.PHONY: re
+re: clean dev
+
+.PHONY: install
+install:
+	npm --prefix=web install
```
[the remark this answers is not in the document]
darrenkuro: I am a little paranoid about the environment and often would do a complete wipeout before making sure to push a PR or deploy or something like
 ```
$(RM) -r backend/node_modules/ backend/dist/
$(RM) -r web/node_modules/ web/dist/
```
and after which I have to manually install both in subdirectories if I don't have a make target (so it's just I'm being lazy really)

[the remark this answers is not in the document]
Comment (cubernetes) [reply Makefile:57]: Alrighty, sounds good :D. Nonetheless, a bit confused why --prefix doesn't work for prettier... here it obviously works

File: README.md (hunk 2 above)
Comment (cubernetes) [reply README.md:36]: Lol you have helix? Amazing xD. But yeah markdown standard wants you to have short lines, my editor also complains a lot.

File: Makefile [hunk 13]
```diff
@@ -1,17 +1,72 @@
...
+	$(D) ps -aq | xargs -r docker stop || true
+	$(D) ps -aq | xargs -r docker rm || true
+	$(D) images -q | xargs -r docker rmi || true
+	$(D) volume prune -f || true
+	$(D) network prune -f || true
+	$(D) system prune -a --volumes -f || true
+	$(DC) down -v || true
+
+.PHONY: fclean
+fclean: deepclean
```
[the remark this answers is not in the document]
darrenkuro: > Maybe this should be merged with the deepclean target? fclean is normal for the other 42 projects, but it's kinda weird here, because it's doing so much (not only cleaning files, but purging all volumes, all images, all networks, all containers, and build cache). Actually, there should be a confirmation asking if you REALLY want to do this, since other developers might have images of their own, and this target just indiscriminately deletes those. Other developers might get angry at us.

Hahaha that's very true, but I feel like Docker is meant to be rebuild. Every container anyone uses should have their Dockerfile configured, I imagine. For some reason I don't trust the cache system for Docker (also the graphic version is kind of ugly too) I feel like I've ran into many unforeseen results because something wasn't cleaned throughly enough (for inception and other things I ran in docker occasionally). But that's probably because it's not a tool I am super used to. I do have my own script to do my own docker wiping though. This target was actually added by @bejasch I think, maybe he can weight in as well. Overall I do agree that wiping docker should be on your own and not be included in the build process.

backend/Dockerfile
cubernetes: Interesting...

backend/Dockerfile
cubernetes: Yup

Makefile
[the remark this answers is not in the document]
cubernetes: Yeah deepclean was added by him, we can keep it for now and remove it before we set the project as finished. I will add the interactive warning tho, so people know what will happen when they execute this target

darrenkuro:
All should be good except Makefile, especially the formatters, as mentioned was sort of a clutch, just documenting the command that worked for me. Feel free to decide how to implement auto-format and/or validate a commit confirm to the format however you see fit in another branch. Let me know if there's anything else before we can merge this! (Idk if it's too many changes in this PR, loading here in gh is getting slower and slower for some reason lol)

cubernetes:
Alright! Will address the formatting in another branch

cubernetes:
I will run the project once more locally and then come back here

Makefile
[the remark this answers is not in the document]
cubernetes: Can mark as resolved

Makefile
```diff
@@ -1,17 +1,72 @@
...
+EXTS_PATTERN := $(subst $(SPACE),|,$(EXTS))  # Convert "ts json" → "ts|json"
+PRETTIER := npx --prefix=backend prettier --write
+
+.PHONY: format format-staged
+format:
+	$(PRETTIER) $(addprefix "**/*.",$(EXTS))
+
+# Only formats staged files, seems as a popular thing to have and seems to be working
+format-staged:
+	git diff --name-only --cached --diff-filter=ACM | grep -E '\.($(EXTS_PATTERN))$$' | xargs $(PRETTIER)
```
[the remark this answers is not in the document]
cubernetes: Will do myself in another branch for the pre-commit hooks

Makefile
cubernetes: Any comments on this one? Otherwise, can be marked as resolved

cubernetes: Alrighty, awesome

cubernetes (approved):
After the addressed changes, looks good to me and works on my machine! :D

PR #17, opened 2025-03-09
Hotfix/improve deployment even more

cubernetes:
- Thought about this for a bit too long, but yeah commits should be self-explanatory

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -20,6 +20,9 @@ services:
                 - path: "./web/Caddyfile"
                   action: rebuild
                   target: "/etc/caddy/Caddyfile"
+        env_file:
+          - .env
+          - config.env
 
     backend:
         build:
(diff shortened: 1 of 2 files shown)


compose.yaml
```diff
@@ -37,13 +34,11 @@ services:
             - DB_PATH=${DB_PATH:-}
             - JWT_SECRET=${JWT_SECRET:-}
         volumes:
-            - ./backend/drizzle:/app/drizzle
-        env_file:
-          - .env
-          - config.env
+            - "drizzle:/app/drizzle/"
```
darrenkuro, line 37: Hmm what happened to env_file here?

cubernetes: Damn

cubernetes: I think I branched off of the wrong branch

cubernetes: Ah no different error. It's a very old compose.yaml. Anyways I'll fix now and push

cubernetes:
compose.yaml should be OK now

darrenkuro (approved):
Not entirely sure how the server is setup but hopefully this works fine haha

PR #18, opened 2025-03-09
Remove quotes in config.env (quoted env vars cause strange bugs)

darrenkuro:
Hopefully this will work on the server, what a bizarre bug lol


cubernetes (approved):
Incredible

PR #19, opened 2025-03-09
Prettier Formatting and Commit Message Pre-Commit Hooks

cubernetes:
- Add pre-commit hook (Format staged files with prettier, abort commit when it found stuff to format)
- Add commit-msg hook (Imperative verb form in titlecase)

### DIFF
--- a/.githooks/commit-msg
+++ b/.githooks/commit-msg
@@ -13,5 +13,5 @@ regex_imperative_verb='^[A-Z][a-z]+ed '
 
 grep -qE -e "$regex_titlecase_verb" -- "$1" || \
 	die "Error: Commit message format is incorrect (must start with titlecase verb). Example: 'Add feature' or 'Update file'"
-grep -qE -e "$regex_imperative_verb" "$1" && \
+! grep -qE -e "$regex_imperative_verb" "$1" || \
 	die "Error: Commit message format is incorrect (use imperative verb form). Example: 'Add feature' or 'Update file'"
--- a/.githooks/commit-msg
+++ b/.githooks/commit-msg
@@ -1,18 +1,17 @@
-#!/usr/bin/env bash
+#!/bin/sh
+# Forces the commit message to be in a certain format
 
-red=$'\033[31m'
-green=$'\033[32m'
-rst=$'\033[m'
+set -eu
+
+die () {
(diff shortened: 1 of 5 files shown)


.githooks/commit-msg
```diff
@@ -0,0 +1,17 @@
...
+die () {
+	printf '\033[31mError: %s\033[m\n' "${1:-"Script $0 exited with unknown error"}"
+	exit 1
+}
+
+regex_titlecase_verb='^[A-Z][a-z]+ '
+regex_imperative_verb='^[A-Z][a-z]+ed '
+
+grep -qE -e "$regex_titlecase_verb" -- "$1" || \
+	die "Error: Commit message format is incorrect (must start with titlecase verb). Example: 'Add feature' or 'Update file'"
```
darrenkuro, line 15: duplicate "Error" here since die() already has it, so it should just be `die "Commit message..."`?

cubernetes: True, I'll fix and push right away

cubernetes: Fixed

darrenkuro:
I am not entirely clear about what happens when it finds stuff to format... I was envisioning when attempting to commit, it would first format all local files, then commit formatted version automatically. Currently it seems to work like will commit if all staged pass the formatter (nothing to change) and will prevent the commit if it does execute a format. And I am not entirely clear on what should be the next steps here. It should be fine since ideally that scenario would rarely if ever come up, especially for those of us who are lazy and have format-on-save enabled.

darrenkuro (approved):

cubernetes:
It's bad practice to let the formatter auto stage files:
https://github.com/pre-commit/pre-commit/issues/806
(I trust Anthony Sottile, I watched probably more than a 100 of his videos on YT)

Ideally, you would want to run git diff to see the formatting changes, or better yet, run tests to make sure it still works after formatting (bit overkill most of the times). As noted in the link, if you satisfied with the changes (`git diff`), you can just run `git add -u && !!`

cubernetes:
Quote:
> In my mind this is one of the worst things that lint-staged does and suggests -- hooks are very frequently not perfect and magically changing what's being committed should not be taken lightly.

PR #20, opened 2025-03-09
Feature/websockets - basic websocket game logic communication

bejasch:
Added to class WebsocketService to start a new active connection and a game session.
On client side, keyevent is tracked and sent as move + up/down/stop if there is a change in the event.
The websocket message is then sent to the new GameEngine class to handle game logic on the server, which broadcasts back to player(s).

This is only a basic set-up which is missing many steps, but movement is tracked and the game loop is handled server side. :-)

### DIFF
--- a/backend/src/controllers/websocket.controller.ts
+++ b/backend/src/controllers/websocket.controller.ts
@@ -1,7 +1,6 @@
 import type { FastifyRequest } from "fastify";
 import { WebSocket } from "ws";
 import WebsocketService from "../services/websocket.service";
-// import type { Player, GameSession, GameState } from "../models/types";
 
 // import { validateId } from "../utils/validator";
 
@@ -34,11 +33,6 @@ export default class WebsocketController {
 
             // Retrieve the updated game state
             gameState = this.websocketService.getGameState(gameId);
-
-            if (gameState) {
-                // Broadcast the updated state to the client
-                conn.send(JSON.stringify(gameState));
-            }
         });
 
         conn.on("ping", () => {
--- a/backend/src/controllers/websocket.controller.ts
+++ b/backend/src/controllers/websocket.controller.ts
@@ -1,44 +1,44 @@
 import type { FastifyRequest } from "fastify";
 import { WebSocket } from "ws";
 import WebsocketService from "../services/websocket.service";
+// import type { Player, GameSession, GameState } from "../models/types";
+
 // import { validateId } from "../utils/validator";
 
-export type Player = {
-    socket: WebSocket;
-    playerId: string;
-};
-
-export type GameState = {
-    ballPosition: { x: number; y: number };
-    score: { player1: number; player2: number };
-    paddlePosition: { [playerId: string]: { y: number } }; // Optional, depending on your logic
-};
-
-export type GameSession = {
-    gameId: string;
-    players: { [playerId: string]: Player };
-    state: GameState;
-};
-
 export default class WebsocketController {
+
     constructor(private readonly websocketService: WebsocketService) {}
 
     handleConnection(conn: WebSocket, request: FastifyRequest) {
-        // Testing
-        let gameState = {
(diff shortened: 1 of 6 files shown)


darrenkuro:
I think we are planning to do something with CLI later right? That together with AI opponent, so I suggest move the pong engine to its own micro-service independent of the Fastify backend. The other good thing about that would be we can use whatever language we want to for the game engine (I think?)

But I am not very sure especially since I am still not clear on how the CLI client should end up working or what we are expected to do with the AI opponent; was just thinking about it and conversing with chatGPT ([key points](https://chatgpt.com/share/67ce273b-ccbc-8006-9568-d1955105f62d)).

cubernetes:
Good points touched upon by @darrenkuro, that's a later milestone for sure, so I personally will approve this PR, but yeah in the future, since @yas0nia might want to do the AI part (preferably in Python since that's their strong suite), we might want to revisit this issue.

Although I'm not sure actually if it might be forbidden according to the subject, what do they mean by 'write the backend in fastify'. Does that rule out any other language for any other potential task? Any thoughts?

cubernetes (approved):
Not a TS pro, but it definitely works on my machine :D

darrenkuro:
> Good points touched upon by @darrenkuro, that's a later milestone for sure, so I personally will approve this PR, but yeah in the future, since @yas0nia might want to do the AI part (preferably in Python since that's their strong suite), we might want to revisit this issue.
> 
> Although I'm not sure actually if it might be forbidden according to the subject, what do they mean by 'write the backend in fastify'. Does that rule out any other language for any other potential task? Any thoughts?

I actually have no idea if it's permitted by the subject. I'm sure it's fine if you separate them (since micro services is actually a module too), and of course not all microservices would be using fastify. The part I am really not sure is that whether other of these services need to be in typescript only or any preference is fair game (I'd imagine forcing people to work on AI in typescript is a bit strange but who knows). Have you seen how the other more "completed" transcendence repos handle this?

cubernetes:
Yeah, true if microservices is a possibility then shoehorning people to use fastify is a bit weird.

Other repos probably just use python throughout, since the previous subject told us to use django in the backend, but I'll double check

cubernetes:
mc-putchar, python+django:
https://github.com/mc-putchar/ft_transcendence/blob/main/pong/game/game_manager.py

demostanis, rust:
https://github.com/demostanis/ft_transcendence/blob/main/backend/pong-server/src/input.rs

darrenkuro:
Personally I think writing this in rust would be so much fun haha. I am still trying to figure out the best architectural design here for how the games should be implemented. That might take some time though. Should I just approve for now since this is more of a proof of concept (I believe?) and we can discuss and potentially change it later? My only worry is that if we don't decide on an architect early and had got change it, there would be a lot more rewriting/refactoring later on.

cubernetes:
Let's wait for @bejasch take on this :D. Don't approve yet, you got a valid point with making good decisions early which will prevent unnecessary work. After @bejasch's go I would approve after which he can merge

bejasch:
Good to discuss at this point, though this merge should only give the basic idea and experience, and also a possibility for John to work and interact with the websocket.

For the future - as we also aimed for the **developement of a CLI** - it makes sense to talk about **microservices** (which can also blow things up).

My main understanding in this case would be a setup like this:

[ Web Client ]
                        HTTPS -->   [Frontend (reverse_proxy)]    --->  [ REST API / Auth (JWT) ]
[ CLI Client ]                                                ↓                                           ↓  
                                                            WebSocket         --->     [ Game Engine Server ]


All traffic goes through Caddy, which routes:
1. to authenticate via REST API
2. to interact via websockets with the game engine server

cubernetes:
Sounds good, what do you say @darrenkuro, ready to approve? Should the next PR be about separating the concerns?

darrenkuro (approved):

cubernetes:
@bejasch This can be merged now :D

PR #21, opened 2025-03-10
Makefile target for local TLS support

cubernetes:
- Add a bit of shell trickery and another Makefile target so that caddy listens on http and https and doesn't autoredirect http to https (which it does by default)
- Some formatting things

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -2,4 +2,4 @@
 
 # This file should only contains secrets. Makes continuous deployment easier
 
-JWT_SECRET='changeme'
+JWT_SECRET=changeme
--- a/.env.example
+++ b/.env.example
@@ -1,3 +1,5 @@
+# Important: Never put quotes anywhere: https://github.com/docker/cli/issues/3630
+
 # This file should only contains secrets. Makes continuous deployment easier
 
 JWT_SECRET='changeme'
--- a/Makefile
+++ b/Makefile
@@ -15,20 +15,20 @@ check-env:
 dev: check-env
 	[ -n "$(DEV_HTTP_PORT)" ] && \
 	[ -n "$(DEV_HTTPS_PORT)" ] && \
-	[ -n "$(DEV_DOMAINS)" ] && \
+	[ -n "$(DEV_SITES)" ] && \
 	HTTP_PORT=$(DEV_HTTP_PORT) \
 	HTTPS_PORT=$(DEV_HTTPS_PORT) \
-	DOMAINS=$(DEV_DOMAINS) \
+	SITES=$(DEV_SITES) \
 	$(DC) up -d --build
 
 .PHONY: prod
 prod: check-env
 	[ -n "$(PROD_HTTP_PORT)" ] && \
 	[ -n "$(PROD_HTTPS_PORT)" ] && \
-	[ -n "$(PROD_DOMAINS)" ] && \
+	[ -n "$(PROD_SITES)" ] && \
 	HTTP_PORT=$(PROD_HTTP_PORT) \
 	HTTPS_PORT=$(PROD_HTTPS_PORT) \
-	DOMAINS=$(PROD_DOMAINS) \
+	SITES=$(PROD_SITES) \
 	$(DC) up -d --build
 
 .PHONY: down
--- a/Makefile
+++ b/Makefile
@@ -31,6 +31,15 @@ dev-tls: check-env
 	SITES="$$(printf %s "$(DEV_SITES)" | sed 's/http\(:\/\/[^[:space:]]*\) /http\1 https\1 /g')" \
 	CADDY_EXTRA_GLOBAL_DIRECTIVES="$$(printf 'auto_https disable_redirects')" \
 	$(DC) up -d --build
+	# Explanation of the SITES variable:
+	# - $(DEV_SITES) is a space-separated list of sites (i.e. "http://localhost https://whatever http://127.0.0.1")
+	# - then we're printing that string and pipe it through sed to do some transormation of that strings
+	# - namely, what we want is do turn every http sites into a https site (for tls), BUT ALSO keep the old http site
+	# - Therefore, the example string from above would turn into this:
+	# - "http://localhost https://localhost https://whatever http://127.0.0.1 https://127.0.0.1"
+	# - note that both http sites got turned into https sites, but the one https site that was already there was not downgraded to http, this is intentional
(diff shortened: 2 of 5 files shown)


darrenkuro:
Could you maybe just explain shortly what these are doing?
```
SITES="$$(printf %s "$(DEV_SITES)" | sed 's/http\(:\/\/[^[:space:]]*\) /http\1 https\1 /g')" \
 CADDY_EXTRA_GLOBAL_DIRECTIVES="$$(printf 'auto_https disable_redirects')" \
``` 

And what I can do to test whether it's working correctly? Everything works fine ofc, I am just not sure what changed in local dev environment from before.

cubernetes:
Sure! I'll make a commit with a comment so it's transparent for everyone persistently

cubernetes:
Added comments inside the Makefile :D

cubernetes:
Ah and you can test it by setting this in config.env
```sh
SITES=http://localhost http://127.0.0.1 http://127.0.0.2
```

And then try to connect to all these sites via https instead, it should work (of course, the warning you will get in the browser is inevitable, but it should connect and everything)

cubernetes:
Of course, in the dev environment, you have to connect to https://localhost:8443 or https://127.0.0.2:8443

darrenkuro:
Yeah that was what I thought it should do, but using https://localhost:8080 gives me an error still, that was why I asked to see if I am misunderstanding something.. (Also I meant to explain here, probably shouldn't have long detailed comments to clutter files like Makefile? It's so nicely written though lol maybe put in a section for README if you want?)

cubernetes:
Yeah it got a bit long, maybe I'll do a wiki entry. There's not really any place for it in the codebase honestly haha

And also you're connecting to the wrong address, you have to specify the TLS port, 8443

darrenkuro:
Ohh yeah you are right. It still gives me an error though but it changed to something else that might due to the local browser security stuff? I feel like it looks familiar but I can't find the button that say "continue to connect" or something..
Screenshot 2025-03-10 at 1 37 31 AM

cubernetes:
You might have to try it in an incognito window, the browsers are sometimes quite aggressive in caching this

darrenkuro:
Vs when I do make prod: 
Screenshot 2025-03-10 at 2 00 31 AM

cubernetes:
Hm okay so I also had some weird issues. I made some commits and will push them now. But those problems weren't really related to your issue, for you it should already work, especially with localhost :/

cubernetes:
Let me know how it is working now :D

cubernetes:
And just to be absolutely sure, you are running `make dev-tls` right?

darrenkuro:
> And just to be absolutely sure, you are running `make dev-tls` right?

I actually did have an "a-ha" moment for "wait it's a different target", since I've been using `make re` (any reason why it wouldn't just replace dev, like, do they have separate functions?) But unfortunately (or fortunately) it still doesn't work when I do use the correct target.

darrenkuro:
I don't know if this is a related at all but just to make sure, you know that your sed for http:// work only when there are trailing spaces? 
```
printf %s "http://localhost" | sed 's/http\(:\/\/[^[:space:]]*\) /http\1 https\1 /g'
http://localhost

printf %s "http://localhost " | sed 's/http\(:\/\/[^[:space:]]*\) /http\1 https\1 /g'
http://localhost https://localhost 
```

cubernetes:
Damn, that's weird...
So you're right actually, at this point, we can combine with targets since they've grown to be quite similar haha. As for `sed`, look at the most recent version, I've fixed that.

As to why it's still not working, wtf. On my machine, these work when I'm on the most recent version of feature/local-tls:
- `make dev-tls`
- `curl -k https://localhost:8443`

darrenkuro:
I do believe that is the issue, adding a space after http://localhost in config.dev would make it work! tested again, without it it doesn't work!

cubernetes:
Ah interesting, then yeah that was the issue dammit. Does it work for you when you're on the most recent version and there's NO space at the end?

darrenkuro:
Nope, just did a pull. Still does not work without trailing spaces.

cubernetes:
What is the sed command you have in the makefile?

darrenkuro:
For some reason after this pull, with trailing spaces it also doesn't work anymore. (Definitely worked before I did a pull.) Current version: 
```
	SITES="$$(printf %s "$(DEV_SITES)" | sed 's/http\(:\/\/[^[:space:]]*\)\( \|$$\)/http\1 https\1 /g')" \
	CADDY_EXTRA_GLOBAL_DIRECTIVES="$$(printf 'auto_https disable_redirects')" \
	CADDY_EXTRA_SITE_DIRECTIVES="$$(printf 'tls internal')" \
```

cubernetes:
How about now :D, it's greatly simplified. If it still doesn't work, can you look at the caddy logs and run curl with -v (verbose) flag?

cubernetes:
Oh wow magic

cubernetes:
Great! Can you also test websockets please?
wscat --no-check --connect wss://localhost:8443

darrenkuro (dismissed):

cubernetes:
I'll wait for @bejasch's review before merging since this is also interesting for his work regarding websockets I guess

darrenkuro:
> Great! Can you also test websockets please? wscat --no-check --connect wss://localhost:8443/ws

Works fine as well.

bejasch:
Works with wscat. For routing websockets (back and forth) through Caddy, it should do the same, but couldn't verify since there is no action available with websockets. This may be visible with the other PR.

In the Makefile, if '.env' doesn't exist, it should call the cp from 
```
check-env:
	@test -f .env || cp .env.example .env
```
But since it didn't existed at include (line 6), Makefile does not run.
Add a '-' before the include (silently ignores missing file):
```
-include .env
```

bejasch (dismissed):
Works with wscat. For routing websockets (back and forth) through Caddy, it should do the same, but couldn't verify since there is no action available with websockets. This may be visible with the other PR. I approve and we'll see.

cubernetes:
Awesome! I'll merge after you've reviewed the last commit that I've pushed which is the off topic thing you've mentioned. It was not a trivial fix, actually, but see yourself:
https://github.com/cubernetes/ft-transcendence/pull/21/commits/a266c7f2d10be4478513e2373228a0d934f7da39

@darrenkuro @bejasch

bejasch (approved):

darrenkuro (approved):

PR #22, opened 2025-03-10
Some cleanups

cubernetes:
- Clean up a few things in the repo
- Also adds some minor convenience features not worth their own PR
- Improve `fclean`, `clean` and `deepclean` targets
- Add more stuff to `README`
- Add `:8080` default port to `Caddyfile`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,9 +74,9 @@ deepclean:
 
 .PHONY: re
 re: clean
-	$(MAKE) dev
+	$(MAKE)
 
-.PHONY: install
-install:
+.PHONY: npm-install
+npm-install:
 	npm --prefix=web install
 	npm --prefix=backend install
--- a/Makefile
+++ b/Makefile
@@ -9,39 +9,39 @@ include config.env
 
 .PHONY: check-env
 check-env:
-	@test -e .env || { \
-		printf '\033[33m%s \033[42;30m%s\033[m\033[33m\n%s\n%s\n%s\n%s\n%s\033[m' \
-			"Warning: .env doesn't exist, trying to run" "cp .env.example .env" \
-			"Warning: Since the file didn't exist, no environment variables from" \
-			"Warning: that file will be included/exposed inside this Makefile." \
-			"Warning: It is STRONGLY recommended you abort this step, manually" \
-			"Warning: copy the file, adjust its contents, and then run make again." \
+	@test -e .env || {        \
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -9,39 +9,41 @@ include config.env
...
 	}
 
 .PHONY: dev
 dev: check-env
-	[ -n "$(DEV_HTTP_PORT)" ] && \
-	[ -n "$(DEV_HTTPS_PORT)" ] && \
-	[ -n "$(DEV_DOMAINS)" ] && \
-	HTTP_PORT="$(DEV_HTTP_PORT)" \
+	[ -n "$(DEV_HTTP_PORT)" ]  &&  \
+	[ -n "$(DEV_HTTPS_PORT)" ] &&  \
```
darrenkuro, line 27: why the && in between the 2 ports and the domain?

web/Caddyfile
```diff
@@ -34,4 +34,3 @@
 		reverse_proxy backend:3000
```
darrenkuro, line 34: Also since they are next to each other... here the backend port is hard coded to 3000 so just wondering whether changing the BACKEND_PORT in config.env actually work or no? I am not sure if docker compose is forwarding it correctly

Makefile
cubernetes: What do you mean? I also don't really recognize that diff, but it should look like this:
```sh
dev: check-env
    [ -n "$(DEV_HTTP_PORT)" ]  &&  \
    [ -n "$(DEV_HTTPS_PORT)" ] &&  \
    [ -n "$(DEV_DOMAINS)" ]    &&  \
    HTTP_PORT="$(DEV_HTTP_PORT)"   \
    HTTPS_PORT="$(DEV_HTTPS_PORT)" \
    DOMAINS="$(DEV_DOMAINS)"       \
    ...
```

And the first three lines of the target are to verify that those 3 variables are set and not empty, and the three lines after that such that those 3 variables are in the environment of the `docker compose up` command

web/Caddyfile
cubernetes: Good catch! I fixed it and pushed

Makefile
darrenkuro: oh yeah ok the diff is fixing spacing

darrenkuro (approved):

bejasch (approved):

Issue #2, opened 2025-02-26
Test-Issue to test discord webhook integration


cubernetes:
just to check if discord webhook integration is working

cubernetes:
it works! what about comments?

PR #23, opened 2025-03-11
Feature/api+test

darrenkuro:
**To be reviewed only for now, I'll rebase and solve conflicts after the other PRs are merged.**

- Changed from a technical layer structure (services/ models/ routes) to a modular structure (each module has its own directory and has service, model, route, etc.) This complies much more with the Fastify architectural philosophy of encapsulating logic per domain and register via plugin so it'd scale more cleanly and easier to test. 
- Following the plugin-context model, rewrote classes into Fastify plugin wrapper and now style is much more uniform and codebase is cleaner looking. Among the modules, user is probably the most complete for its general calls. Please check that to see if the general structure looks good. `[/backend/src/user]` If it seems acceptable I can complete the rest pretty easily since they all follow the exact same structure. And the rest of whatever route is need depends on what sort of calls other services are making. Shouldn't be too difficult under the current structure.
- Added Zod as validation library instead of ajv which is the default for Fastify. Zod seems to be very popular these days and is quite nice to use. Used a general function to hook the validation call before any data is passed to the controller so that controller will always receive correct data. It is a little complicated looking at the moment since I wanted to combine all validation into one call (validating header, params, body, etc.) so it's not the easiest general function to read. But I think it's a pretty neat solution. [`/backend/src/utils/zod-validate.ts`]
- Added tap for unit testing and some sample test files. 
- Now passing the development/production env from the very top (Makefile -> Docker compose -> Docker build stage) to change build (remove dead code when in production for compiling on the backend, minify code output for example) and code behavior when process is running. 
- Expanded prettier to format files that it supports. If any file or types of file wants to be excluded we can also do that. And making some minor style changes as I go. ( I've read up so much on the debate on backtick vs double vs single quote and still couldn't decide on what's the best solution. :/ )
- Removed unnecessary structures (remnants of NestJS), nothing needs to be class in Fastify, you can just keeping add plugins and hooks. Entry point (index.ts) is basically the old app instance. Also wrapped config together to be loaded first. And now the app cleans up safely (ish?), all onClose hooks will be called to do any clean up jobs. (I.e. closing database connection.)

### Note from Timo
- This branch also includes some commits from me which were from a branch called `refactor/minor-web-stuff` which basically did some improvement of the stuff in the web folder: See here for diff: https://github.com/cubernetes/ft-transcendence/compare/62621c740d2d80dcd88b6c039fc66f456f7109e5...52323d7a015b3c4e22ef53fd5362503b35b7b613

### Note from Timo #2
- This branch now also includes some commits from john which were from a branch called `feature/babylonjs_scene` which added a simple babylon scene and didn't have much conflict with the other code


darrenkuro:
Also, there is mock data named password for a fake user for testing user module that's caught by security check. Not sure what's the best practice here since it's no real, maybe there's some config that can ignore certain files or folders? Not sure can get around with mocking fake secrets in unit testing.

cubernetes:
Wow! Comprehensive! I looked a little bit at it, look good to me I guess :D. I added an exception for the credentials, so now GitGuardian passes it

backend/src/utils/config.ts
```diff
@@ -0,0 +1,88 @@
...
+import { z, ZodError } from "zod";
+
+const configSchema = z.object({
+    BACKEND_PORT: z.coerce
+        .number({
+            required_error: "cannot be empty",
+            invalid_type_error: "must be a number",
+        })
+        .int("must be an integer")
+        .min(1024, "must be greater than 1024") // Avoid privileged / reversed
```
cubernetes, line 15: Should say "must be greater than 1023"

backend/src/utils/config.ts
```diff
@@ -0,0 +1,88 @@
...
+        .max(65535, "must be less than 65535"),
```
cubernetes, line 16: Should say "must be less than 65536"

backend/src/auth/auth.plugin.ts
```diff
@@ -1,5 +1,5 @@
 import fp from "fastify-plugin";
-import { createAuthService } from "./auth.service";
+import { createAuthService } from "./auth.service.ts";
```
cubernetes, line 2: Why is `.ts` needed here? I thought it's implicit?

cubernetes: Lol I should read the commit message first haha

backend/Dockerfile
```diff
@@ -2,6 +2,9 @@ FROM node:23.9.0-alpine3.21 AS builder
 
 WORKDIR /app
 
+# Docker compose env needs to be manually set at build stage
+ENV NODE_ENV=${NODE_ENV}
```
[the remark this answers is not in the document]
darrenkuro: it is needed for build script, esbuild.config.js, which is what npm run build calls. Apparently after some testing and researching, the compose.yaml only takes care of running environment at the end container, so seems to be a multistage build problem. I am not exactly sure how it work, but basically the problem I was trying to solve is without this line, npm run build (esbuild.config.js doesn't get a process.env.NODE_ENV)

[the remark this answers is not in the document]
cubernetes: Interesting, well, it doesn't hurt :D

cubernetes:
When do you think this is mergable?

cubernetes:
And who should need to review this? I looked at it, but should @bejasch know about these changes as well?

darrenkuro:
Okay, made some final changes and this should be ready to merge now. 

- Unify parentheses for single line if function, arrow function syntax. 
- Refactor websocket module to comply with the current architecture.
- Move deleting local files (nodule_modules etc.) to deep clean, since I don't want those deleted when I do a make re which I prefer. And local files shouldn't have anything to do with docker containers anyways. But deep clean actually is more like a docker clean, maybe make it two separate targets?
- Add config to backend for frontend to call to get. Currently I just port it to call in function. And I separated it from the rest, which, after some thoughts I think belong to the frontend, since those are basically rendering/styling factors and only relevant to the 3d part of the game which belongs entirely to the frontend. Later on maybe can just fetch once at some top level (when game starts?) instead of in place (createGameObjects), since some calculations may require those as well. Repetitive call to static response is probably bad idea. Also, most frontend stuff probably needs to eventually end up being async. 
- Notice some issues currently (maybe I'll raise in issues tab to keep track to be implemented?): 1) When moving away from game tab, game is still ongoing, can hear the sound effects; not sure if it drop connection correctly. Probably just not implemented, just a note. 2) Log any info only in dev mode so user don't get any. Not sure how frontend separate the two modes when build so didn't touch anything there for now. 3) "Clean" probably shouldn't be an npm script but handled by Makefile, it's probably there for working in that directory, maybe we can delete it at the end. 
- @cubernetes how does the server get .env? Since I update the requirement for JWT_SECRET it might not start on the server after merge, I can also turn the requirement off for now.

cubernetes:
Awesome! Great work. Regarding the server, I already have a long JWT_SECRET there, so that's fine.
Regarding the config, yeah I thought that the frontend just fetched the config once and then has it for the entire lifetime of the site.

But now that I think of it, it's not really clean to have frontend configuration hard coded in the backend code, right? Not sure how to solve this

darrenkuro:
> Awesome! Great work. Regarding the server, I already have a long JWT_SECRET there, so that's fine. Regarding the config, yeah I thought that the frontend just fetched the config once and then has it for the entire lifetime of the site.
> 
> But now that I think of it, it's not really clean to have frontend configuration hard coded in the backend code, right? Not sure how to solve this

I think frontend will handle and host the frontend config. What it is getting from the backend are the configuration for the game like speed of the ball, size of stuff, etc. (that are shared in some ways since wherever the game engine lies it needs these data.) That's why I only moved a small chunk of the config, the rest, imo, should live on frontend (since it's about rendering and has nothing to do with the game logic.) It should be pulled once every time a game is created, which makes it possible to have different values for whatever reason (difficulty, accessibility, settings, etc.)

darrenkuro:
> > Awesome! Great work. Regarding the server, I already have a long JWT_SECRET there, so that's fine. Regarding the config, yeah I thought that the frontend just fetched the config once and then has it for the entire lifetime of the site.
> > But now that I think of it, it's not really clean to have frontend configuration hard coded in the backend code, right? Not sure how to solve this
> 
> I think frontend will handle and host the frontend config. What it is getting from the backend are the configuration for the game like speed of the ball, size of stuff, etc. (that are shared in some ways since wherever the game engine lies it needs these data.) That's why I only moved a small chunk of the config, the rest, imo, should live on frontend (since it's about rendering and has nothing to do with the game logic.) It should be pulled once every time a game is created, which makes it possible to have different values for whatever reason (difficulty, accessibility, settings, etc.)

cubernetes:
I see! Makes sense

I'll review now

.githooks/pre-commit
```diff
@@ -37,4 +37,4 @@ format_staged () (
 	fi
 )
 
-format_staged ts json
+format_staged ts js json css html md yaml
```
cubernetes, line 40: Maybe add yml as well? Just in case

backend/Dockerfile
cubernetes, line 6: ### Note for Timo (don't make a commit for this)
- remove `=`

.github/workflows/deploy-production.yaml
```diff
@@ -1,59 +1,59 @@
...
+                     set -Eeuxo pipefail ;                                                                            \
+                     cd ${{ secrets.PRODUCTION_WORK_DIR }} ;                                                          \
+                     git stash --include-untracked || : ; : 'git stash may fail (if it does, prolly docker related' ; \
+                     git checkout --force ${{ secrets.PRODUCTION_BRANCH }} ;                                          \
+                     git pull --rebase --force ${{ secrets.REMOTE_NAME }} ${{ secrets.PRODUCTION_BRANCH }} ;          \
+                     make clean ;                                                                                     \
+                     make prod"
 
-      - name: Cleanup
-        run: rm --recursive --force ~/.ssh
```
cubernetes, line 59: ### Note for Timo (don't make a commit for this)
- use short-options for GNU utils

Makefile
```diff
@@ -29,6 +29,7 @@ dev: check-env
 	HTTP_PORT="$(DEV_HTTP_PORT)"   \
 	HTTPS_PORT="$(DEV_HTTPS_PORT)" \
 	DOMAINS="$(DEV_DOMAINS)"       \
+	NODE_ENV="development"         \
 	CADDY_EXTRA_GLOBAL_DIRECTIVES="$$(printf %b "$(DEV_CADDY_EXTRA_GLOBAL_DIRECTIVES)")" \
 	CADDY_EXTRA_SITE_DIRECTIVES="$$(printf %b "$(DEV_CADDY_EXTRA_SITE_DIRECTIVES)")"     \
 	$(DC) up --build -d
```
cubernetes, line 35: ### Note for Timo (don't make a commit for this)
- use long-options

backend/Dockerfile
```diff
@@ -2,6 +2,9 @@ FROM node:23.9.0-alpine3.21 AS builder
...
+
 COPY ./package*.json ./
 RUN npm install
 COPY . .
```
cubernetes, line 10: ### Note for Timo (don't make a commit for this)
- use trailing slash uniformly for all directories

web/Dockerfile
```diff
@@ -36,7 +36,7 @@ COPY                ./Caddyfile /etc/caddy/Caddyfile
 RUN rm -rf /srv/ && mkdir -p /srv/.local/share/caddy/
 
 # Copy built site from Node.js build stage
-COPY --from=web-builder --chown=caddy:caddy /app/dist/ /app/public/ /srv/
```
cubernetes, line 39: Wait, how come /app/public isn't copied anymore?

cubernetes (approved):

web/Dockerfile
darrenkuro: John moved it to src and webpack is coping it when it's doing its bundling.

cubernetes: Ah perfect

cubernetes:
Skimmed it, most of the stuff I've already reviewed. I think it should be merged now :D, this PR has been hanging around for too long

PR #24, opened 2025-03-11
Refactor minor web stuff

cubernetes:
- Just me reviewing the entire web folder and familiarizing myself with typescript a little, this is in preparation to make hot-reloading work, but for that, I need to know a bit about the ergonomics of working with the code
- Note that this is a merge request into `feature/api+test`, not `main`
# Minor Functionality changes
- Redirect to home when invalid route is encountered
- Redirect to home when header is clicked
- Simplify type signature of routes object by making all `create{Home,Game,Profile}Page` function async (at not cost, since `await` would wrap them in a Promise anyway later on)

### DIFF
--- a/web/src/scripts/router/Router.ts
+++ b/web/src/scripts/router/Router.ts
@@ -15,11 +15,17 @@ export function createRouter(container: HTMLElement): void {
         // Get the route from the URL hash (without the #)
         const route = window.location.hash.slice(1);
 
+        // Redirect to home upon invalid route
+        if (!(route in routes)) {
+            window.location.href = "/#home";
+            return;
+        }
+
         // Clear the container
         container.innerHTML = "";
 
         // Render the appropriate page
-        const createPage = routes[route] || routes[""];
+        const createPage = routes[route];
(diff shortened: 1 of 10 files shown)


web/src/scripts/router/Router.ts
```diff
@@ -16,11 +15,17 @@ export function createRouter(container: HTMLElement): void {
...
+        if (!(route in routes)) {
+            window.location.href = "#home";
+            return;
+        }
+
         // Clear the container
         container.innerHTML = "";
 
         // Render the appropriate page
-        const createPage = routes[route] || routes[""];
```
cubernetes, line 23: Guaranteed to exist because of line 19

web/src/scripts/router/Router.ts
```diff
@@ -16,11 +15,17 @@ export function createRouter(container: HTMLElement): void {
...
+        const createPage = routes[route];
```
cubernetes, line 28: Guaranteed to exist because of line 19

darrenkuro (approved):
Looks fine!

PR #33, opened 2025-03-16
Add watch for index.html and assets, browser hot-reload on change

darrenkuro:
- Hot reload should now be working fine for index.html and any file change, addition, removal in assets directory. Likely can’t handle subdirectory, but it should work as long as everything is in root under `src/public/assets`.
- Browser now auto-refresh on any change made on the frontend.
- Rebuild on top-level change (`esbuild.config.js` and `reload-server.js` for frontend.)

### DIFF
--- a/app/esbuild.config.js
+++ b/app/esbuild.config.js
@@ -13,7 +13,8 @@ const context = await esbuild.context({
     minify: isProd, // Minify in production
     sourcemap: !isProd, // Disable sourcemaps in production
     define: {
-        "process.env.WATCH": watch ? "1" : "0",
+        "process.env.WATCH": JSON.stringify(watch ? "1" : "0"),
+        "process.env.NODE_ENV": JSON.stringify(isProd ? "production" : "development"),
     },
 });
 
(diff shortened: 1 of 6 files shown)


darrenkuro:
> minor and not related to this one here: WDYT about creating a separate logging.ts for the logging config?

Oh yeah sorry forgot to mention it's done. Currently @ 'backend/src/utils/logger.ts'.

cubernetes (approved):
Looks all good! Rly awesome. Little nitpick because you're using ws://localhost:35729 instead of a generic solution, but it's fine since it's for local development only :D. Maybe a bit problematic that this port is always exposed in the app container, even in production mode, but eh whatever

cubernetes:
Can be merged

darrenkuro:
> Looks all good! Rly awesome. Little nitpick because you're using ws://localhost:35729 instead of a generic solution, but it's fine since it's for local development only :D. Maybe a bit problematic that this port is always exposed in the app container, even in production mode, but eh whatever

Yeah I was thinking whether there's a way to toggle the expose in docker compose conditionally but there probably isn't a way afaik unless you have two compose (like compose.prod.yaml n compose.dev.yaml)? Didn't feel like an exposed port here would be particularly worth the effort but that might be more standard for production that's much more concerned about security? 

Would love to figure out whether there's other options possible (maybe can be done in line at with docker compose cli on Makefile level?) but will merge for now.

PR #36, opened 2025-03-18
Use yaml anchors for various things in compose.yaml

cubernetes:
- Uses a nice feature I got to know from [a friend of mine](https://github.com/mustaphazorgati/server-applications/blob/main/zorgati_cloud/compose.yml), what do you think? Reduces config duplication a bit in the yaml


cubernetes:
Sure :D, I'll write a post in the random channel

PR #37, opened 2025-03-18
Update/app rename

OwnerOfJK:
1. Change app directory name from `app` into `frontend`.
2. Adjust sound effects from. (I tested it and I hope its a lot better but lkm lol)


cubernetes (approved):
Works! :D

PR #39, opened 2025-03-20
ELK core functionality

yas0nia:
Simple ELK setup configured to "catch" ft-transcendence- logs
Kibana logs preview is available under `localhost:5601/app/discover `

to run, use: `docker compose -f compose.elk.yaml up -d    `

From the subject:

✅Deploy Elasticsearch to efficiently store and index log data, ensuring it is easily
searchable and accessible.
✅Configure Logstash to collect, process, and transform log data from various
sources (various? not really?), sending it to Elasticsearch.
✅Set up Kibana for visualizing log data, creating dashboards (not really), and generating
insights (not really) from log events.
❌ Define data retention and archiving policies to manage log data storage e"ec-
tively.
❌Implement security measures to protect log data and access to the ELK stack
components.

I'll address all that is not done now in the following PRs

Example:

Screenshot 2025-03-21 at 00 31 46


darrenkuro:
The containers are up and working but I couldn't make it to show any data. Regardless, just a couple questions:

- Is there a specific reason why it's on a separate docker compose file and not together with the other containers? (To reduce build time maybe?) But in the end they are supposed to be in one docker compose, correct?
- I think maybe there are some mistakes with npm, since top level wasn't supposed to be have package*.json (backend/frontend each is its own.)
- This is more of a Mac issue for whoever it may concern, port 5000 is often in use and you'll have to turn off airplay receiver for that port to be open so docker compose runs. [(ref)](https://nono.ma/port-5000-used-by-control-center-in-macos-controlce)

backend/src/utils/logger.ts
```diff
@@ -36,14 +36,53 @@ const formatGameState = (state: GameState) => {
...
+                options: {
+                    mode: "tcp",
+                    address: process.env.LOGSTASH_HOST || "logstash",
+                    port: parseInt(process.env.LOGSTASH_PORT || "5000"),
+                    reconnectTimeout: 1000,
+                }
+            }
+        ]
+    },
+    serializers: { err: formatError, gameState: formatGameState },
```
darrenkuro, line 87: It's probably unlikely that production needs any serializer (esp the error one considering the log level) but probably doesn't hurt to have them

cubernetes:
Regarding the 2 compose files, let's keep it like this for now (maybe add a makefile target soon), but for the future, maybe [compose profiles](https://docs.docker.com/compose/how-tos/profiles/) might be a clean solution to this? Having a similar issue with hashicorp vault, so I'll do smth like that as well probably

cubernetes:
OK, works nicely, didn't test thoroughly, but we can merge :D

cubernetes:
I'll remove the requirement to have signed commits until this is merged, but please sign the commits for future commits :D (here's a 17 second how-to video: https://www.youtube.com/watch?v=IrK83nKi8HA)

cubernetes (approved):

PR #42, opened 2025-03-21
Feature/api docs

darrenkuro:
- Main thing is that swagger ui is available at `/api/docs/` or `/api/docs`. Changing some underlying structure to reflect schemas more comprehensively on this page. 

- Unify response type when succeeding or failing.
- Using Fastify hooks to handle errors instead of try-catch block on every level, only changed (parts of) user module since not exactly committed to this structure. It should be cleaner in the long run.
- Fix various small issues in Makefile, compose.yaml, etc. See commit msgs.


cubernetes:
> swagger ui is available at /api/docs or /api/docs
Typo? The two slugs seem to be identical

cubernetes (approved):
Works well :D, can be merged

darrenkuro:
> > swagger ui is available at /api/docs or /api/docs
> 
> Typo? The two slugs seem to be identical

Oh yeah one is with a trailing slash.

PR #43, opened 2025-03-22
upd package lock bc I forgor

yas0nia:
so since im new to npm and I keep forgetting to update the pckg lock if there r conflicts and I resolve them manually in the ui I added a new PR workflow check to check if the files r in sync


darrenkuro (approved):
Looks good

PR #45, opened 2025-03-23
Feature/user+login

darrenkuro:
Proof of concept that the login, registration, JWT, protected routes with Zod for backend system works as expected. 
(Frontend component used only as an example interface.)

Will do some clean up after reviewing the core ideas/flows here are correct.


cubernetes:
I love that it immediately logs you in when you register. Far too few websites actually do that. Played around a bit and works great! Nitpicks would be that the login page is still available after login, although I'd expect that it'd be hidden (But I guess the way it's currently is actually not bad in any way, I don't know why other website don't do this, maybe it's confusing?).

Another nitpick is that you can't test the /api/users/me route unless you somehow get the token (either console.log or getting it from localStorage.token) and then manually doing the curl requests. Bit cumbersome, but not important.

Can be merged I think :D

cubernetes (approved):

darrenkuro:
> I love that it immediately logs you in when you register. Far too few websites actually do that. Played around a bit and works great! Nitpicks would be that the login page is still available after login, although I'd expect that it'd be hidden (But I guess the way it's currently is actually not bad in any way, I don't know why other website don't do this, maybe it's confusing?).
> 
> Another nitpick is that you can't test the /api/users/me route unless you somehow get the token (either console.log or getting it from localStorage.token) and then manually doing the curl requests. Bit cumbersome, but not important.
> 
> Can be merged I think :D

Right, I'd expect it to be hidden too. I am sort of expecting the frontend to go through some design changes and/or refactor later (I also don't think login should be its own page for instance but rather a component inside another page/view). So rather the frontend component is just to show that backend functionality works, it's clearly rather messy to have all business logic in one file, and the fetch in test.ts (Just trying to contain the changes in limited amounts of files in frontend.) 
The issue is I really suck at design from scratch but once there is, say, a detailed Figma file or copying visual logic from an existing site, things become much easier for me. Maybe I'll try harder to find some inspirations after I clean up the backend a bit.

cubernetes:
Yeah, a fleshed out figma would be a dream haha, maybe we should hire someone through fiverr for a few bucks to make a kick-ass frontend design for us 😝. But yeah for now it's good as it is no worries, it's great to see that it is working well

PR #47, opened 2025-03-24
Feature/blockchain

OwnerOfJK:
Add blockchain integration to the Game Page. This change allows the user to:
1. Connect their blockchain wallet.
2. Request game data stored in a smart contract based on the gameID.
3. Trigger and sign a transaction to post game data onchain.

This also includes the smart contract written in solidity.

What is not part of this PR:
1. Dynamic blockchain transaction based on current game data. (This is mainly due to the lobby / game session not yet fully working).
2. Current UI/UX design is a work in progress.


frontend/src/contracts/contracts.history.sol
```diff
@@ -0,0 +1,38 @@
+// SPDX-License-Identifier: MIT
```
cubernetes, line 1: Let's not put License identifiers in code, the CC0 doesn't require that so it can be removed :)

OwnerOfJK: Okay, this is fine with me. Its just common practice for smart contracts to add an open-source license to the contract itself.

cubernetes (approved):
Not tested, but looks good in general :)

PR #49, opened 2025-03-24
Fix wallet connect bug

OwnerOfJK:
Fixed an issue where the game wouldn't launch if the user does not have an ethereum wallet installed.

How?

Added a try and catch and error checks to ensure wallet is initialised. If not an error is printed.


cubernetes (approved):
lgtm!

PR #50, opened 2025-03-24
Set up vault service with lots of automation

cubernetes:
First of all, this is so weird. So many things are happening that wouldn't happen in an actual environment, leading to questionable design choices for the sake of automation. To list a few
- You wouldn't extract the unseal keys and immediately use them to unseal the vault (you'd rather PGP-encrypt them and then every owner would individually use them to unseal the vault)
- You wouldn't inject vault tokens themselves directly to the services (either via environment variables or better bind mounts, like it is at the moment) (you'd rather create [auth methods](https://developer.hashicorp.com/vault/docs/auth) and let the services authenticate so that they get an auth token back)
- You wouldn't use HTTP (you'd rather create your own [CA and setup PKI](https://developer.hashicorp.com/vault/tutorials/pki/pki-engine). client certificates would also be securely distributed to the services)
- you wouldn't save the root token and the unseal keys in the vault container (you'd rather revoke the root token completely and let the auth methods do their job (issuing short lived tokens) and regarding unseal keys, you'd just not shut down vault that often. every time vault is shut down, it MUST be unsealed using the unseal keys, which should be a manual task)
- you wouldn't put vault in the same compose project (you'd rather have as a completely different, orthogonal service that is long-running, afaik)

OK, with that out of the way, here's the
## Key features
1. New `vault` service (reachable in `ft-transcendence-net`) which runs the vault server in production mode with a simple file backend and without High Availability (HA) (see simple `vault/config.hcl`)
2. Automatically unseals itself upon first startup (if the volume is empty)
3. Automatically populates the key-value (kv) secrets engine with the secrets provided in `vault/env.json`
4. Secrets in `vault/env.json` that can be generated can be specified using a simple custom [mustache template](https://en.wikipedia.org/wiki/Mustache_(template_system)) syntax, implemented via a small custom python script (`vault/replace_json_templates.py`), i.e. `{{digit:10}}` to create a sequence of 10 random digits, or `{{alnum:64}}` for the JWT_SECRET. *All secrets should preferably use this*, unless they come from a 3rd party service and cannot be generated (like OAuth/Google Sign-In)
5. Creates vault tokens (current policy is `root`, later it will have proper policies and `TTL`s) and writes them to a special bind-mounted file (currently only for `backend` service), to which other services have access to. Other services should truncate this file after use, for security.
6. If `SAVE_UNSEAL_KEYS` and `SAVE_ROOT_TOKEN` environment variables are `1`, it will save those secrets in files in a folder
7. Upon restart and if the `SAVE_UNSEAL_KEYS` and `SAVE_ROOT_TOKEN` environment variables are `1`, it will read those secrets back and unseal the vault, revoke all vault tokens (TODO), and then create new tokens and write them to files as described in 5.
8. Doesn't reinitialize upon restart by checking if the volume is not empty
9. Add healthchecks and depends_on rules in Dockerfile/compose, that check if vault is unsealed and running, which is important for all services that depend on vault (like `backend` atm)
10. Extensive logging: Specify via the LOGLEVEL environment variable (debug, info, warn, error) in `compose.yaml`


backend/src/utils/config.ts
```diff
@@ -24,10 +24,27 @@ export const appOpts: FastifyServerOptions = {
     logger: process.env.NODE_ENV === "production" ? prodLoggerConfig : devLoggerConfig,
 };
 
+export const readVaultSecret = async (vaultToken: string, apiBaseAddr: string, path: string, key: string) => {
```
cubernetes, line 27: @darrenkuro  This and following lines is the part that you can definitely take a look at :'D, it's working, but probably a bit clunky

darrenkuro: I would probably move this to index.ts, because well the issue that I can think off the top of my head is testing environment, that secret (& the whole vault container) won't be available.

darrenkuro: I can do that though since I changed some other minor stuff too

darrenkuro (approved):
It works; didn't look deep into the scripts for the vault... Only question that I had was to the function of having the secrets directory kept as an empty folder, seems like Makefile is taking care of ensuring that directory exists whenever compose up already, unless it's also doing something else?

backend/src/utils/config.ts
cubernetes: Ah, this is necessary so the tests have access to it right? Because normally, if you can, you would try to avoid putting secrets into environment variables at all cost, but I guess there's no clean way around this if you want to have them in the test code as well, right?

cubernetes:
I see, no fair point. The reason the makefile has these checks is to ensure the a regular file exists at those locations, because otherwise the volume mounts create directories no matter what (very annoying...). But yeah, I would probably keep the directory anyways, so people don't accidentally create a `secrets` directory accidentally and have it suddenly removed...

backend/src/utils/config.ts
darrenkuro: I meant, it wouldn't actually be an environment variable since it's only available in memory at run time and can't be accessed outside, so as far as I'm concerned that'd just be like any other variable... unless I am missing something. Reason solely is consistency and not having to change much, possibly to overwrite it as a potential option. Tests really don't have much to it because I'd imagine in unit testing it'd mock it's own secret and wouldn't use "the real one," haven't gone as far as integration test, but it feels viable as well.

darrenkuro:
Could you explain what you mean by `so people don't accidentally create a secrets directory accidentally and have it suddenly removed`? It just seems that with the .gitignore in `secrets` you went out of your way to make sure the repo contains that empty folder when it could be local and let it ignored on the top level; because every time compose ran in make, the folder and the file is guaranteed to be there?

backend/src/utils/config.ts
cubernetes: Well, I think technically it still counts as an environment variable, but not as an *initial* environment variable, which means other processes cannot just run `xargs -0n1 < /proc/<PID>/environ` to dump the secret. From what I've researched, you need to execute ptrace system calls to access the current environment of another process, which is possibly disabled in docker (unless added back with --cap-add or --privileged).

In other words, it's really fine, so yeah, when you refactor, you can just do that :D

cubernetes:
Hm, fair enough, that makes sense. I'll add it to .gitignore and remove it, this way it's clear it should be created. Thanks!

cubernetes:
I added another commit because I completely forgot to unseal the vault when restarting the project lol. Had to really change a few things for that. Should be still fine though. Gonna wait for @yas0nia's review before I merge, want two people looking at this

backend/src/utils/config.ts
darrenkuro: Cool cool. 

If you have ptrace privileges you have access to everything in memory right? In which case nothing would be safe.

cubernetes: Depends, if it's an environment variable, and you have ptrace permissions, then it's trivial to read it. However, if it's just in memory and not an environment variable, it's a little bit more difficult since it's not as easy as just printing `char **environ`. But technically you still have access.

A way to mitigate this would be to not run PID 1 in the backend container as root, but as an unprivileged user (see caddy container), but that's maybe for another time as it's not suuper crucial. Or use bitnami docker images ([this project uses them](https://github.com/kichkiro/ft_transcendence/tree/main)) where processes or non-root by default, but again, maybe for another time

cubernetes:
Lol fr tho. I will make another PR and add a few important bits and pieces from the description to the README

Makefile
```diff
@@ -55,8 +57,14 @@ define wait-progress
...
 
+.PHONY: ensure-secret-files
+ensure-secret-files:
+	@mkdir -p ./secrets/
+	@$(RM) -r $(VAULT_TOKEN_EXCHANGE_FILES)
+	@touch $(VAULT_TOKEN_EXCHANGE_FILES)
+
 .PHONY: dev-old-compose
-dev-old-compose: check-env clean-app-volumes
+dev-old-compose: check-env clean-app-volumes ensure-secret-files
```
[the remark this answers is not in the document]
cubernetes: Fair, I'll make a small comment explaining why this target can be useful

compose.yaml
```diff
@@ -10,10 +10,43 @@ x-env-common: &env-common
     BACKEND_PORT: ${BACKEND_PORT:-3000}
 
 x-env-files: &env-files
-    - .env
     - config.env
 
 services:
+    vault:
```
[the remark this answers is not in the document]
cubernetes: No preference from my side, I'll put it on the bottom

vault/entrypoint.sh
```diff
@@ -0,0 +1,389 @@
...
+			;;
+		(INFO) printf 2
+			;;
+		(WARN) printf 1
+			;;
+		(ERROR) printf 0
+			;;
+		(*) 	printf 0
+			;;
+	esac
```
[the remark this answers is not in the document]
cubernetes: Bash is one hell of a cursed language eh?

vault/replace_json_templates.py
```diff
@@ -0,0 +1,102 @@
+#!/usr/bin/env python3
+# Author: https://github.com/cubernetes
+# License: Public Domain
+#
+# This unix filter replaces simple mustache templates in json strings. Example:
```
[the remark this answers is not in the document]
cubernetes: Since we will have multiple secrets, most of which can (should) be generated, it's nice to have a way to specify the secrets without hardcoding them. Take [this](https://github.com/kichkiro/ft_transcendence/blob/main/project/.env.json) for example, they have 7 hardcoded secrets for which they have placeholder values (not convenient...). By using this declarative syntax, we can easily specify secrets in source control in a secure way without hardcoding them.

Of course, mustache syntax is just a choice I had to make, since it's relatively straightforward (and of course, I wanted to have a little bit of fun :') )

If a developer needs access to the secrets, they can look them up in vault (I'll add to README how to do it)

vault/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+
+RUN apk add --no-cache jq python3
+
+COPY ./entrypoint.sh /entrypoint.sh
+COPY ./replace_json_templates.py /replace_json_templates.py
+COPY ./healthcheck.sh /healthcheck.sh
+COPY ./config.hcl /vault/config/config.hcl
+COPY ./env.json /vault/config/env.json
+
+HEALTHCHECK --interval=10s --timeout=2s --start-period=8s --start-interval=1s --retries=3 CMD ["/healthcheck.sh"]
```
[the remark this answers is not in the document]
cubernetes: Marking as resolved

vault/replace_json_templates.py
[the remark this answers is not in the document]
cubernetes: Marking as resolved as well

PR #57, opened 2025-03-27
Add a CLI for PONG (standalone)

bejasch:
Added the CLI for PONG which doesn't cover local games yet, but connects over websocket.

Run "make install" and start it from /cli with "npm start" for now.

I'm highly interested in feedback referring the audio!


darrenkuro:
From what I can tell aplay only works in Linux so on macOS error msg as follows. Not a big deal potentially if it'll be dockerized properly later, but play-sound potentially is a better library as it'll detect host os and tries available players. 
```
node:events:485
      throw er; // Unhandled 'error' event
      ^

Error: spawn aplay ENOENT
    at ChildProcess._handle.onexit (node:internal/child_process:286:19)
    at onErrorNT (node:internal/child_process:484:16)
    at process.processTicksAndRejections (node:internal/process/task_queues:90:21)
Emitted 'error' event on ChildProcess instance at:
    at ChildProcess._handle.onexit (node:internal/child_process:292:12)
    at onErrorNT (node:internal/child_process:484:16)
    at process.processTicksAndRejections (node:internal/process/task_queues:90:21) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'spawn aplay',
  path: 'aplay',
  spawnargs: [
    '-q',
    '--buffer-size=96000',
    '--period-size=48000',
    'src/content/menu.wav'
  ]
}

Node.js v23.10.0
```

cli/README.md
```diff
@@ -0,0 +1,59 @@
+
```
darrenkuro, line 1: Can't do inline comment for binary file so will just do it here: I believe this file should not be here `backend/data/database.sqlite`

Makefile
```diff
@@ -124,6 +124,7 @@ re: clean
 install:
 	npm --prefix=frontend install
 	npm --prefix=backend install
+	npm --prefix=cli install
```
darrenkuro, line 127: Make target `fclean` should add the rm for cli as well

cli/src/GameRendering.ts
```diff
@@ -0,0 +1,260 @@
+import { Vec2D, ICLIGameState, GameConfig, FieldConfig } from "./game.types";
+import { userOptions } from "./options.js";
+import { audioManager } from "./audio";
+
+let TERM_WIDTH = 160;
+let TERM_HEIGHT = 40;
+let GAME_CONFIG: GameConfig | null = null;
+let FIELD_CONFIG: FieldConfig | null = null;
+let FIELD_BUFFER: string[][] | null = null;
```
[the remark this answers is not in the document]
bejasch: Oh, good to know 👍

cli/README.md
bejasch: The play-sound was used in my first version, but on my system the sound stuttered after a time, while also having huge latencies [but it may be a cause of me using WSL].
I will try more different options and optimise for our use case.

Makefile
bejasch: True

bejasch:
- Resizing terminal window will make a mess of things! It is up to the user to choose a right resolution or change terminal settings!
- Key release doesn't work! It is not tracked in stdin-stream. There is no good solution I've found to undergo. So the solution was set to a "STOP"-key which is by default space, but can also be set in the options-menu.
- Centralising shared types is a good idea for the project-package! But in the end, as a standalone, it should be copied into the cli-sources.

darrenkuro (approved):

bejasch:
The KeyRelease simulation was discussed with your research. The problem which is not solvable on the coding site IMHO is, that the key repeat is set in the system settings. After the first key stroke, there is always a little pause until retriggering the key. This leads to bad behaviors, meaning no usable results.

darrenkuro:
Right, I see the problem. It seems that it is possible to bypass the system delay by simulating your own key repeat with a debounce/timer, and would be easier with a library that normalize keypresses potentially like blessed. But definitely not something trivial, and it won't be 100% perfect. I guess if anyone is interested could play around with it and see later.

cubernetes:
Unfortunately tho, @bejasch is right, it's [really impossible](https://blog.robertelder.org/detect-keyup-event-linux-terminal/). At least the way we want it. Even if we are super restrictive and say that it shall only run in a Linux environment with X11, even then it's impossible, unless we forward the X11 socket to docker. But even if we do that, it's a non-solution since it will not work on any Linux system that uses Wayland (or variants thereof) and it will not work on MacOS or Windows.

Let's just say, the only satisfying solution would be to:
1. Implement branching logic that distinguishing between Windows, MacOS, and Linux (X11 and Wayland)
2. Depending on the 4 major possibilities, use vastly different strategies with Docker
 1. Linux X11: Forward the X11 socket to docker
 2. Linux Wayland: No idea, it's way harder, since it's [up to the compositor to decide](https://news.ycombinator.com/item?id=27184132) what key events to send, so good luck
 3. MacOS: Impossible, requires an Apple approved driver. The only one currently approved seems to be Karabiner (well known MacOS remap tool), so you have to hook into their system. [Source](https://www.reddit.com/r/C_Programming/comments/ve1l2t/comment/icoi6zf/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button)
 4. Windows: Don't get me started...

And that's all you got without root. But even with root, it's not easy. With root, you goal would be to write a keylogger, basically. For Linux (independent of windowing system, hurray), you would do something similar to what evtest, showkey -s, or X (internally) do. See the very first link I have at the top. For MacOS, no idea, it might be straight up impossible. Same for Windows, when I run `sudo showkey -s` in WSL2, it shows absolutely nothing.

cubernetes:
The only solutions that seem viable within our timeframe are
1. Have a dedicated stop key (e.g. space)
2. Press the same key again to signal a stop
3. Tailor it to the school computers and therefore rely on X11. It will not work on @darrenkuro's, @OwnerOfJK, or @yas0nia's Macbooks and also not on my Windows Laptop. It would require us to install a VM or dual boot Linux with X11.

darrenkuro:
Yeah I think we can agree that it is impossible to detect **true** keyup event on a terminal environment, there are potentially somewhat satisfying simulation mechanisms (to somewhat reliably bypass key repeat issue) that it is possible to employ that works well for, say, 95% of cases, or is that misinformation as well? Blessed doesn't provide a key release event ofc (since I don't think that concept exists in terminal at all) but from what I gathered it does provide some background support that helps with simulation to make it marginally more reliable. 

I could be completely wrong though, it was like a mere hour-long research of a topic that I barely know anything about.

cubernetes:
Well, thanks to your skepticism Darren, we know have a very well working solution! I've coded up a Python PoC that works flawlessly and is almost indistinguishable from an actual solution that recognizes KeyRelease events! If you reduce the initial keyboard repeat delay, it feels more and more natural.

Even better so, when you quickly press the same key twice, the second press event is recognized as a release event instead, which is perfect for our use-case! (You'd want the paddle to rather stop than to continue moving, at least it makes sense to me, but I haven't _felt_ it yet so can't really know).

This system works so well that you can even use it over SSH (or Docker fwiw). The way it works it pretty much your penultimate comment (I think, correct me if I'm wrong). Every system initial keyboard delay, and a separate delay when the keys start repeating. Once you know these values (in milliseconds), you can more or less infer when a key has stopped being pressed down. And the thing you weren't sure about, whether this works well or not, well: It works very well. If you spam two keys *as fast as you can*, the delay between two key presses is roughly 100ms. If you're lucky, 80ms or less.

However, when two keys repeat on windows, the delay is only 30ms, way lower. This means a KeyRelease can be recognized 2-3 times faster than you can double press a single key.

The only downside for this is the needed initial calibration, but that's doable. In fact, here's all the code! Have fun with it, especially @bejasch. I'm all ears for questions, I tried to comment each line as much as possible, but I think after doing webserv, it should not be too hard to understand ;).

https://github.com/cubernetes/kbd_release_hack

darrenkuro:
I was honestly just curious, find this issue somewhat interesting, and figured after some quick research that it could be somewhat viable, and therefore as aforementioned "maybe will try it out later." Leave it to Timo to have a prototype within a couple of hours 😅 such a go-getter lol

bejasch:
Just to clarify what I was talking about, saying 
> This leads to bad behaviors, meaning no usable results.
I meant, that there is no possibility to get a key release in between the first keypress and the second trigger, where system auto repeats. On my system, the key repeat lies at ~40 ms, while the repeat start happens at ~500 ms. But **I need to get the key release event maybe at around 150 ms**.

cubernetes:
That is technically correct, but it doesn't mean the result will be unusable. Pressing the key and immediately releasing it would lead to the paddle moving in the desired direction for longer than you wanted. But that's only a disadvantage if the paddle is too small (in fact, given the initial repeat delay and velocity of the paddle, you can calculate the minimum length needed to make the paddle still usuable). The longer the delay, the worse the user experience of course. And the paddle moving against your will is only 1 of 3 cases. The other 2 cases are pressing the opposite key (instantly move the other direction) and pressing the same key again (almost instantly stop moving (40ms delay), at least with my PoC that I linked above).

Not that it's of any priority, but as Darren said, I think it's an interesting challenge to solve and I think the behavior from the python script is the best possible way to handle this without relying on a second key

bejasch:
I tried this alternative and it doesn't really feel responsive to me.
Having a stop key on the other hand gives exact precission. 

So I would leave it like this for now and merge this version. Later, it should be reviewed anyway, because of changing logic for the AI opponent and tournament system.

cubernetes:
Ah, I probably should've communicated this, I didn't mean to require this to be in this PR, I just wanted to solve a hard problem. With 500ms delay, I can understand that this is probably quite bad of an experience nonetheless. You can merge this :D

cubernetes:
Although personally I like pressing the same key twice much more than a stop key, it's one less thing in my head to juggle while playing the game. Fortunately, this can be added alongside the stop key at a later stage ^^

PR #65, opened 2025-03-29
Feature/hotreload support vault

cubernetes:
Adds hotreload support while also using vault, by creating ephemeral tokens ONLY in production. Also fixes a small logging issue in backend because I forgot to approve copilot's suggestions in the last PR.

### DIFF
--- a/backend/src/core/config/config.plugin.ts
+++ b/backend/src/core/config/config.plugin.ts
@@ -27,7 +27,9 @@ const readVaultOnce = async (path: string) => {
 			throw new Error(`Request to vault API FAILED:\n${e}`);
 		});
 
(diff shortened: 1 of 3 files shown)


cubernetes:
@darrenkuro how does it look? Should work now :D. I'm going to sleep, so if you approve of this today already you can merge it immediately instead of me. It's just a small change so no worries

darrenkuro:
Volumes had to be cleaned then it works which makes sense, but during the process I realized every time this compose is called two unnamed volumes are created, likely due to `"./.secrets/backend_vault_token:/run/secrets/backend_vault_token"` (educated guess..) Is this expected?

cubernetes:
I'll investigate

cubernetes:
@yas0nia conflict was my fault lol

@darrenkuro good guess, I would've had the same, but it's actually something completely different haha. Created another PR that clarifies and cleans it up, #66

PR #66, opened 2025-03-30
Refactor/cleanup volumes

cubernetes:
- Caddy had an unused VOLUME (declared *only* in the Dockerfile -> therefore anonymous) -> removed it
- The hashicorp/vault image specifies a /vault/logs volume, but we don't specify one in compose.yaml. I added it there
- Also some style changes in compose.yaml

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -108,6 +108,7 @@ cleandb: clean
 .PHONY: cleanvault
 cleanvault: clean
 	$(D) volume rm --force ft-transcendence_vault
+	$(D) volume rm --force ft-transcendence_vault-secrets
 
 .PHONY: fclean
 fclean: clean cleandb cleanvault
--- a/Makefile
+++ b/Makefile
@@ -93,6 +93,8 @@ down:
 .PHONY: clean-app-volumes
 clean-app-volumes: down
 	$(D) volume rm --force ft-transcendence_frontend
+	$(D) volume rm --force ft-transcendence_vault-logs
+	$(D) volume rm --force ft-transcendence_elasticsearch-data
 
 .PHONY: clean
 clean:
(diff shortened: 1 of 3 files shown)


darrenkuro:
What happened with the Make failure for GitAction on main?

cubernetes:
Haha see PR #67

darrenkuro:
For Makefile target clean vault, should the additional volumes also be removed?

cubernetes:
Good idea, I'll add it (you only mean the vault-logs and elasticsearch-data volumes right?)

cubernetes:
Merge?

darrenkuro:
vault-logs and vault secret? I'm not sure if elastic search-data volume has anything to do with the cleanvault target. Not exactly sure when to clean what volumes though..

cubernetes:
Would you want to re-generate all secrets when you run `make re`?

darrenkuro (approved):
The `clean` targets seem a little confusing but I am kind of unsure, but that can be cleaned up later. As far as the volumes go this pr fix it fine.

cubernetes:
Yeah haha I'm also not sure what should clean what and so on. I'm gonna make one more commit that clean vault-secrets when running `make cleanvault`

darrenkuro:
> Would you want to re-generate all secrets when you run `make re`?

Potentially? I feel like clean vault seems to indicate that it'll come back fresh. Otherwise why cleaning the vault anyway? And they should probably be all grouped together into something like a clean-app-volumes (when you do, you have a fresh copy of db, vault, elasticsearch, basically everything except for caddy?) Maybe the frontend has its own target and gets triggered everytime, and everything else get cleaned together with a clean-volumes target? Just that sometimes when things are wrong (i.e. a pr is not working properly and the reason is a persisted volume but you are not sure which one); manually figuring out the issue is a bit of pain than having a catch all target. Idk.. it's kind of whatever, not very important, we'll figure that out later.

cubernetes:
Made another PR #68 to clean up the Makefile a little

PR #68, opened 2025-03-30
Refactor/Makefile

cubernetes:
# Key Features
- Creates semantically more sound clean targets
- Puts them in a different Makefile (to be included)
- Also extract auxiliary stuff into another Makefile

# Ancillary Changes
- Move env.json into repo root
- Rename vault volume
- Remove check-env dependency for prod target
- Only run `make down` in github actions workflow


Makefile
```diff
@@ -1,72 +1,31 @@
...
-	@mkdir -p ./.secrets/
-	@$(RM) -r $(VAULT_TOKEN_EXCHANGE_FILES)
-	@touch $(VAULT_TOKEN_EXCHANGE_FILES)
+# Helper Makefile
+include Makefile.clean
+include Makefile.aux
 
-# This target is needed for legacy docker compose versions where there's no `--watch' flag for `docker compose up'.
-# Use the `dev' target instead.
+# This target is needed for legacy docker compose versions where there is no `--watch' flag for `docker compose up'.
```
darrenkuro, line 24: Mismatched quotes? I'm pretty sure there are ticks paired with single quotes?

Makefile
```diff
@@ -1,72 +1,31 @@
...
+# The `dev' target is preferred.
 .PHONY: dev-old-compose
-dev-old-compose: check-env clean-app-volumes ensure-secret-files
+dev-old-compose: check-env
+	$(MAKE) clean-app-volumes ensure-secret-files
```
darrenkuro, line 28: I think this should be `clean-frontend-volume` here now?

Makefile
cubernetes: Common quoting style when quoting code elements, can't give you examples from the top of my head, but I've seen it many times in documentation (for example extensively used in the bash man page)
- https://english.stackexchange.com/questions/17695/any-reference-on-the-usage-of-a-backtick-and-single-quotation-mark-like-this
- https://www.man7.org/linux/man-pages/man1/bash.1.html

.github/workflows/deploy-production.yaml
```diff
@@ -52,7 +52,7 @@ jobs:
                      git stash --include-untracked || : ; : 'git stash may fail (if it does, prolly docker related' ; \
                      git checkout --force ${{ secrets.PRODUCTION_BRANCH }} ;                                          \
                      git pull --rebase --force ${{ secrets.REMOTE_NAME }} ${{ secrets.PRODUCTION_BRANCH }} ;          \
```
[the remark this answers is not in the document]
cubernetes: Nope

Makefile
cubernetes: Correct!

Makefile
cubernetes: Although now that I'm digging into this a bit more, the `' style seems to be used for single characters and ``'' for longer strings, but I've seen ``'' being used much rarer than `'

darrenkuro: Huh interesting, I'll look into it more later but if it's intentional then you can just resolve this

I'd like to be consistent though if possible; so maybe after doing a bit of a research and deciding on this style we will change all the quotes in comments to this styles later?

cubernetes:
Oh lol, sure sure. I'm not 100% sure what scheme you mean exactly, this?:
Makefile.aux -> Makefile.aux.mk
Makefile.clean -> Makefile.clean.mk

In that case, because .mk means make already, I would suggest we rather do
clean.mk
aux.mk
How does that feel? For consisteny, we could even rename the Makefile to makefile (see https://www.gnu.org/software/make/manual/html_node/Makefile-Names.html).

What do you say?

Makefile
cubernetes: We can absolutely do that :D. For ease of collaboration we will probably end up deciding on either plain single or double quotes. The
```
`'
```
style is a little bit opinionated haha, although I feel like relatively common in shell scripts specifically

darrenkuro:
`.mk` does mean Makefile I believe, but I don't really see a big problem with it because since files are sorted often alphabetically, they will be grouped together and that could be more visually pleasing on the side bar. That's the issue in an editor though, I guess if you live in the terminal you'd barely notice lol. Since it'd still be makefile and not makefile.mk, I think I'd rather do Makefile, Makefile.aux.mk, Makefile.clean.mk unless the repetition really bothers you. Let me look into it see if there's any naming conventions. Tbh I didn't even know you could separate Makefile contents like this before.

cubernetes:
Ohh yeah, no the naming makes sense, when I do vim Make<TAB> it's nice to have all 3. Ok I'm gonna rename them

cubernetes:
Gonna have to make one more commit, because as of now none of the clean targets depend on the `down` target, so everything fails lol

darrenkuro:
So apparently the convention for Makefiles per AI (didn't check too much) is that if it's in root folder, they are called `Makefile`, `Makefile.subfilea` `Makefile.subfileb`; and if in a dedicated folder they are called Makefile, and in `/make` you'd have the `subfilea.mk` `subfileb.mk`. I also figured out that I can force the editor (VS code to load Makefile regardless with this in the setting)
```
  "files.associations": {
    "Makefile.*": "makefile"
  }
```

So you could switch back to the old names if you'd like (might be more conventional); but I feel like either way is fine and with `.mk` extension it'd work out of the box for anyone working in vs code.

cubernetes:
Ok haha gonna remove the extension again

darrenkuro (approved):

PR #70, opened 2025-04-01
Feature/game+tournament

### LINKED ISSUE #58: Centralizing shared types
There are some shared types between cli client, frontend, and backend regarding the pong game that I've been thinking to have centralized because they are being changed and improved and instead of copying and pasting and ensure consistency manually, it should probably just be in one place. I've looked into this and probably the best way to do it is just to publish the package into a registry either the public one (it's completely normal practice, just have to make sure no sensitive data, with typing it's probably completely okay) or a private one (GitHub package is a good option?). Not something to worry about just a note, I'll probably make an issue of this later.

_Originally posted by @darrenkuro in https://github.com/cubernetes/ft-transcendence/issues/57#issuecomment-2757834068_

### LINKED ISSUE #59: Import order should be handled by formatter
Import order should be consistent across different file and ideally not handled manually, maybe with [prettier-plugin-sort-import](https://github.com/trivago/prettier-plugin-sort-imports). The issue is in what order they should be, alphabetical doesn't seem to be a very nice choice visually. 
It should probably have some connections to whether it is a type import, if it's from an external package or a local one, order of when it's used, etc.

darrenkuro:
**This PR will break the game section on the app, but it needs to be the foundation of both frontend and cli migration as well as any AI opponent code so after some thoughts and discussions maybe it'd be better to have it merged first.**

- Refactor websocket stuff on the backend so that it is of a clean structure. 
- Implement queueing system and remote game handlers on the backend.
- Move read vault function to comply with test setup for tap, and change error strings to better fit with how the serializer is outputting them.
- Implement a logger on the frontend so that logging level is controllable by production or development mode and the actual clients don't get anything (but errors). @OwnerOfJK I did this a while ago and I kind of forgot about it. Hopefully it's not too much of a conflict to solve with the changes on frontend. I can roll this back and implement it later if there are too many conflicts, lmk.
- Closes #58 . Pong engine and any data required by both server and the client are now in its own (published) package. Currently the package is published under my personal NPM account (I couldn't get the organization one to work for some reason...) Which means changes need to be published by me personally, which could be a pain. Maybe we'll try the Github package registry if it becomes too much of a pain. (Manually copying into node_modules is a simple hack when developing.) [Edit: after some research I believe it's also possible to add collaborators so if if anyone needs write access, I should be able to grant it if you give me your npm username.]
- Closes #59 . Add a prettier plugin to order all the imports automatically so that they are consistent. Formatter is expected to change some old files since I didn't run it on all files, but as long as you have git hooks enabled it shouldn't be a problem.


cubernetes:
Tagged the current version of main
https://github.com/cubernetes/ft-transcendence/releases/tag/game-working
This way, if we need a working version of the game, you can just do
`git checkout game-working`

cubernetes:
(You might have to do `git pull --tags`)

backend/src/index.ts
```diff
@@ -1,12 +1,24 @@
...
 // Fastify server options, cannot be changed once instance is created
 const appOpts: FastifyServerOptions = {
     logger: process.env.NODE_ENV === "production" ? prodLoggerConfig : devLoggerConfig,
 };
 
+// Read vault secrets
+const secrets = await readVaultOnce("secret/data/backend");
+
+if (secrets.isErr()) {
+    console.error(`Fatal error when reading vault secrets: ${secrets.error.message}`);
```
cubernetes, line 15: Should this go to the logger facility instead?

backend/src/utils/vault.ts
```diff
@@ -0,0 +1,35 @@
...
+    let res;
+
+    try {
+        res = await req.json();
+    } catch (error) {
+        return err(new Error("invalid JSON response"));
+    }
+
+    if (!req.ok) {
+        return err(new Error(`API request failed: ${res.errors?.[0] ?? "unknown error"}`));
```
cubernetes, line 27: Quick question for my own clarification, will the error message go to the logger?

cubernetes:
Didn't go through it in great detail, but looks good!

cubernetes:
However, I had to do this to make it run with rootless docker:
```diff
diff --git a/compose.yaml b/compose.yaml
index 4095ab2..2226c65 100644
--- a/compose.yaml
+++ b/compose.yaml
@@ -125,6 +125,7 @@ services:

     logstash:
         image: docker.elastic.co/logstash/logstash:8.17.3
+        user: root
         container_name: logstash
         volumes:
             - "./elk/logstash/pipeline/:/usr/share/logstash/pipeline/"

```

OwnerOfJK (approved):
Went through most changes in particular the pong package and everything on the frontend. Logging changes shouldn't be much of a problem so its fine to keep it. Also reviewed some of the changes in the backend.

Also ran the branch to check it runs the app, and all worked for me. (Weirdly it didn't work for @cubernetes).

Good job, looks very clean.

backend/src/index.ts
darrenkuro: No because app (where logger is from) is not available yet here

backend/src/utils/vault.ts
darrenkuro: It won't, here it's just being returned, no printing is being done.

Actually now that I'm thinking about it; these are not very likely to go through the logger from Fastify since it runs before the app is ever gonna be built. The serializer comment was for from before where it runs inside the tryBuild process in the config plugin. Now it'll just be `Fatal error when reading vault secrets: ${message}`

darrenkuro:
> However, I had to do this to make it run with rootless docker:
> 
> ```diff
> diff --git a/compose.yaml b/compose.yaml
> index 4095ab2..2226c65 100644
> --- a/compose.yaml
> +++ b/compose.yaml
> @@ -125,6 +125,7 @@ services:
> 
>      logstash:
>          image: docker.elastic.co/logstash/logstash:8.17.3
> +        user: root
>          container_name: logstash
>          volumes:
>              - "./elk/logstash/pipeline/:/usr/share/logstash/pipeline/"
> ```

Huh that's weird. I didn't run into this issue either. Also I don't think I changed anything on compose so the current main branch should have the same issue?

backend/src/index.ts
cubernetes: Gotcha!

backend/src/utils/vault.ts
cubernetes: Interesting

cubernetes:
Yeah, it doesn't have to comply with school computer environment/rootless docker. In the worst case we'll run it in a VM. But weirdly, it worked for @OwnerOfJK on the same kind of computer. I'll observe this in the future, made an issue to not forget about it.

cubernetes:
Can be merged I think

cubernetes (approved):

PR #80, opened 2025-04-04
2FA with Time-based OTP

cubernetes:
That took longer than I hoped. Use the script in the `./script/` folder to generate codes (it will take the most recently added user), or scan the QR code to add it to your authenticator.
Styling still has to be done (red buttons, etc.), and the flow is still not quite right (2FA verification page is actually a separate route, which makes things a bit ugly).


cubernetes:
For some routes, I added a window.location.reload(), but this is forbidden because it's not a SPA anymore (the entire point of SPA is to never ever reload).

frontend/src/scripts/router/router.url.ts
```diff
@@ -31,7 +35,9 @@ export const createRouter = (container: HTMLElement): void => {
         }
 
         // Check access for protected routes
-        checkAccess();
+		if (route != "totpVerify") { // TODO: code smell
+			checkAccess();
+		}
```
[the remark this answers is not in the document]
cubernetes: Maybe another tiem

frontend/src/scripts/pages/pages.totpVerify.ts
```diff
@@ -0,0 +1,96 @@
...
+            if (!jwt) {
+                alert("Invalid response from server (JWT missing). Logging out");
+                localStorage.removeItem("token");
+                localStorage.removeItem("username");
+                container.innerHTML = "";
+                window.location.reload();
+            } else {
+                localStorage.setItem("token", jwt);
+                window.location.href = "#setup";
+				window.location.reload(); // TODO: code smell, we shouldn't need to reload the page here
```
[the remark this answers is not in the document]
cubernetes: Maybe another time

backend/src/modules/user/user.controller.ts
```diff
@@ -12,17 +14,121 @@ const registerHandler = async (
...
+    reply: FastifyReply
+): Promise<void> => {
+    const user = await req.server.userService.findById(req.userId);
+
+    if (user.isErr()) {
+        return user.error.send(reply);
+    }
+
+    const secret = speakeasy.generateSecret();
+	const b32secret = secret.base32
```
darrenkuro, line 46: Hmm missing semicolon; is prettier/hooks on correctly?

darrenkuro:
Is there a particular reason to differentiate totp_secret and temporary_totp_secret? Seems like they are one and the same, only difference is the name and reassignment in initial verify. Totp_enabled alone should be enough gate keep, no?

backend/src/modules/user/user.controller.ts
darrenkuro: nvm I assume it's fixed by the other pr haha

cubernetes: Oh, that's sus. It activated I believe. Hm. Anyways, should be fixed by the other PR yes haha

cubernetes:
Thanks, yeah fair, for this one you could write proper tests indeed. The script was just for testing for yourself, for development. Otherwise, it would be a huge hassle to having to add it manually to the 2FA app all the time

cubernetes:
I had only totp_secret in the beginning, but if you look at the documentation for speakeasy, indeed, they demonstrate the usually flow by having a temporary secret. I didn't implement the temporary secret at first, but it bit me later down the line, because think about it. Everytime you go to the 2FA setup page, it will fetch the /api/user/totpSetup endpoint. This endpoint will ALWAYS generate a new secret, necessarily. But you wouldn't want that just going to that page resets your secret right? That's why you need a temporary secret, that may change anytime, but at the same time persist in the database. As soon as you verify the temporary secret on the totpVerifyInitial endpoint, it will permanently save that secret. I think that's the procedure that is easiest to extend/debug/maintain, but surely there's a more compact of doing it by only saving totpSecret and a flag, but yeah that's about it.

cubernetes:
For me it says "Cannot merge at this time". Any idea why? Can you press the merge button?

cubernetes:
Oh why it suddenly changed, for whatever reason

PR #83, opened 2025-04-04
Refactor/run prettier on everything

cubernetes:
Merge #80 first.
Just prettiers everything. Also renamed 2 files from .yml to .yaml because that's the standard.


darrenkuro (approved):
Same as #80

PR #86, opened 2025-04-09
Refactor/frontend

### LINKED ISSUE #71: Implement profiles on compose file for faster startup?
The start up time for compose up is getting very long, not sure why, is it really just logstash being built on java? 

Anyways, looked into it and figured that we can use profiles on the compose file to control which containers are spawn and maybe have elk as a separate profile? 

Just an idea; not sure if it'd be easy to do or even possible with all the interdependencies...

### LINKED ISSUE #78: Docker build warnings
Image

Not sure whether these matters too much but warnings are never good I suppose.

All containers show similar warnings.

Changing to something like this works, but idk if it'll rise other issues.
`CMD ["sh", "-c", "if [ \"$WATCH\" = \"0\" ]; then npm run build && npm start; else npm run start:dev; fi"]`

### LINKED ISSUE #79: Fix schema to show correct error message when username is missing
When verifying a TOTP token, a username must be provided. If it's missing, this error message is returned:
```json
{
  "success": false,
  "error": {
    "code": "VALIDATION_ERROR",
    "message": "Required"
  }
}
```

Instead, it should tell you that the username is missing

### LINKED ISSUE #81: Make it a single-page app again
For some routes, I added a window.location.reload(), but this is forbidden because it's not a SPA anymore (the entire point of SPA is to never ever reload).

_Originally posted by @cubernetes in https://github.com/cubernetes/ft-transcendence/issues/80#issuecomment-2778836518_

### LINKED ISSUE #82: 2FA verification flow is smelly
TOTP verification happens on another route: `#totpVerify`. This is a bit ugly in the code, and it's probably worth it to make it part of the login flow, instead of having another route. To make it work for now, I excluded it form `checkAccess()` in the router in the frontend.

darrenkuro:
**This branch is a working progress but didn't want to hold up too much time, so opening a PR for review. Could be merged soon (by the end of today?) if needed, things are, for the most part functional; albeit still a bit messy and not completely refactored yet especially the UI-related stuff. But that might take another day or so let me know.**

- Please check the `README.md` for frontend for proposed structure, currently structured as such, opinions welcome. UI following tsx naming conventions. Not sure if I like it yet.. 
- Implement as functional approach as much as I can. 
- Add Reactive store for global states.
- Attach stuff globally to window (config, log, because they'll imported everywhere otherwise.)
- Completely refactored game logic to make it more modular. Renderer is completely self contained. It shouldn't concern socket as it doesn't know whether game is local or remote or ai.  Also broke up to multiple files so it is cleaner and stuff are easier to find. Kept the original structure as much as possible but did implement some small changes which structurally I believe made more sense.
- Temporarily moved blockchain and simulation to an archive folder because I am unsure about how to refactor them. 
- UI is the part that still has the most left to refactor. Main goal is to reduce duplicate code and create reusable helper function, and clean duplicate tailwind class as much as possible.
- Fixed some open issues, will be closed automatically due to commit messages. ( Fix #81 , Fix #79 , Fix #78 ) _Huh, I thought commit msg would automatically close them when merged as well but doesn't seem so.._
- Tons of small stuff.. such as adding helper function createEl, so scoped looked more correct, and moved fragment to router to remove unnecessary DOM nesting. Didn't document very well, probably easier to do a checkout pr and just read through the frontend..
- Fix #82 as well I believe (I just realized what you mean @cubernetes , now totpenabled will trigger a modal replacement in the same container so no page change.)
- Fix #71 ....?

**Also, cannot stress enough that this is merely a proposal... don't know much about architectural stuff especially the frontend. Subject to change and discussions, questions, are very much welcome.**


cubernetes:
Hm, I'll take a look. Would be awesome if that would fix it all already!

OwnerOfJK:
Wow huge refactor. First, of all awesome job, thanks for doing this. Sorry for only getting back to you on this now. Took me a little to go through/understand all the code.

General thoughts:
- **Folder Structure:** I like the new directory structure with `global`, `modules`, `ui` and `utils`. Was initially also not sure where the `Paddles.ts` and `LoginForm.ts` best belong. So the `layout` directory fits well.
- [`global/store.ts`](https://github.com/cubernetes/ft-transcendence/pull/86/files#diff-4072de2b690b1b0e7cd164b6909f1994668c98c51f19c00bf293fb7a80d56297): This is honestly pretty amazing. If I understand it correctly, `auth` and `game` both handle their state changes via the `store`? The idea is to handle state globally across the entire app. Is it a little similar how we use `useEffect` in react?
- [`modules/game`](https://github.com/cubernetes/ft-transcendence/pull/86/files#diff-bf5889a0958c9f32e198d02d31b106ba0eab22dd0d22f19e5ea1e2ac2f871ab4): Was initally thinking that the `game` module looks a bit bloated (in particular the `renderer` + `objects`), but after some consideration, the modularisation, will make it easier for us to extend if we decide to. Curious what @bejasch thinks.

Questions:
- [`types.d.ts`](https://github.com/cubernetes/ft-transcendence/pull/86/files#diff-cd1edb9fc2dc5b77e0bc00213f2bf492e65101918d3b7a3a8a07c9b59439d322): I don't get how this works as part of the engine/renderer:
```declare module "@babylonjs/core" {
    interface Engine {
    ...
    }
    
    interface AudioEngineV2 {
    ...
    }
```
- [`utils/dom-helper.ts`](https://github.com/cubernetes/ft-transcendence/pull/86/files#diff-c2941e10282b5381423109236125669810376fa20f1bce97783fb00fea6ceace): I saw your comment above but this looks a bit strange to me. Was your main idea here to get rid of repeated typescript syntax of creating HTML elements?

Additional Comments:
- [`archive/components.blockchain.ts`](https://github.com/cubernetes/ft-transcendence/pull/86/files#diff-5ee36bf6d2e3db2ea71ed7bffbe0d1091e65779026232d88eca228f730187d3f): In order to follow your structure, I am considering to create a `modules/blockchain` dir where I provide the signature for all blockchain/contract interactions. Will also add the current contracts directory in it. The UI for it will be naturally part of the tournament pages.
- [`archive/components.simulation.ts`](https://github.com/cubernetes/ft-transcendence/pull/86/files#diff-d18e5a4d1e1ef6a3f1b598e53a2ee6ccff5cc6a09d31515d8e7bf72844ed9f44): I haven't fully decided what to do with this yet. I might still need it for later. TBD.
- **`ui` refactor:** Based on your previous code, I can support you with the refactoring. Might take me more time than you, so just lkm.

darrenkuro:
> * [`global/store.ts`](https://github.com/cubernetes/ft-transcendence/pull/86/files#diff-4072de2b690b1b0e7cd164b6909f1994668c98c51f19c00bf293fb7a80d56297): This is honestly pretty amazing. If I understand it correctly, `auth` and `game` both handle their state changes via the `store`? The idea is to handle state globally across the entire app. Is it a little similar how we use `useEffect` in react?

Yes, precisely. Due to the extreme popularity of React and similarly conceived frameworks like vue, svelte, etc., the reactive store management of states is by far the most popular and standard way to handle state management so I figured it'd make sense to copy how most people are doing it. Basically it follow the observer pattern which is personally my favorite. (Similar to event listener/emit, store just notify all the subscribers to run a callback whenever the managed state changes.)

> * [`modules/game`](https://github.com/cubernetes/ft-transcendence/pull/86/files#diff-bf5889a0958c9f32e198d02d31b106ba0eab22dd0d22f19e5ea1e2ac2f871ab4): Was initally thinking that the `game` module looks a bit bloated (in particular the `renderer` + `objects`), but after some consideration, the modularisation, will make it easier for us to extend if we decide to. Curious what @bejasch thinks.

I do agree it is a bit bloated. But as a unit (or module) I think renderer is as small as it goes, but I subscribe to the idea that each file shouldn't contain too many lines and pure functions are better approach than OOP design which is easy to enter a level of state back and forth reference hell. If each file is a very clearly defined unit, finding stuff would also be much faster than to search globally. 

And I was also originally planing to make the functional more general; i.e. the idea setup for me personally would be have less than half a page of parameters that I can change, and there's no need to scroll down for factory functions for, say, object creation, etc. But might be too much work and require s bit of a deep understanding of the engine; but might look into it more.
 
> Questions:
> 
> * [`types.d.ts`](https://github.com/cubernetes/ft-transcendence/pull/86/files#diff-cd1edb9fc2dc5b77e0bc00213f2bf492e65101918d3b7a3a8a07c9b59439d322): I don't get how this works as part of the engine/renderer:
> 
> ```
>     interface Engine {
>     ...
>     }
>     
>     interface AudioEngineV2 {
>     ...
>     }
> ```

Basically instead of passing a huge state reference of a bunch objects, I opt out to attach associated object on to the engine directly. Since js objects are prototype based and free of type restriction that other languages have, you can essentially just add key-value pair to any object you have access to and to be used later whenever. Types are defined here so typescript as a language knows that I attached these personally. So basically the idea is whenever a function of something like `initRendererEngine`, it will run a bunch of small functions to generate light, scene, camera, objs, etc. And even though we can probably access them in the engine somewhere nested and buried deep, it is much easier to just attach them at top level, so after a `const engine = initEngine()` We can just do `engine.camera = createCamera(); engine.light = createLight();` etc. Basically to just attach reference onto the instance (think of as a state in a class but there's no restriction.)

Some people really don't like it. I am still on the fence but for a very modular instantiated object like an engine of sort, I think it's probably fine. But kind of a clutch, still trying to think of better ways to manage them.

>  * [`utils/dom-helper.ts`](https://github.com/cubernetes/ft-transcendence/pull/86/files#diff-c2941e10282b5381423109236125669810376fa20f1bce97783fb00fea6ceace): I saw your comment above but this looks a bit strange to me. Was your main idea here to get rid of repeated typescript syntax of creating HTML elements?

 I guess it's a personal preference of scoping. 
```
    const title = document.createElement("h1");
    const a = document.createElement("a");
    a.href = "#home";
    a.textContent = "ft-transcendence";
    a.className = "text-3xl font-bold";
    title.appendChild(a);
``` 

Something like this is obviously fine and preferred by some. But the issue for me is that when I'm skimming through a large body of code, it is very hard to see which lines belong to "one unit" since there's no indentation, it's usually only separated by empty lines and comments, and that is done manually and people aren't very reliable when it comes to stuff like that. Where as with functional wrappers, even without abstracting too much stuff away, it gives us scope with formatter for free, and when you skim through code it is very obvious where does the unit start and end, like this:

```
    const heroEl = createEl(
        "div",
        "absolute top-0 left-0 w-full h-full flex flex-col p-8 rounded-lg text-center",
        { children: [ctaButton] }
    );

    const musicEl = createEl("audio", "", {
        attributes: { src: `${window.cfg.dir.audio}/main.mp3` },
        props: { loop: true, volume: 0.4 },
    });
```

> Additional Comments:
> 
> * [`archive/components.blockchain.ts`](https://github.com/cubernetes/ft-transcendence/pull/86/files#diff-5ee36bf6d2e3db2ea71ed7bffbe0d1091e65779026232d88eca228f730187d3f): In order to follow your structure, I am considering to create a `modules/blockchain` dir where I provide the signature for all blockchain/contract interactions. Will also add the current contracts directory in it. The UI for it will be naturally part of the tournament pages.

I don't really know how blockchain stuff works but I think that makes sense. Feels like most of it are business layer logic and not much UI. 

> * [`archive/components.simulation.ts`](https://github.com/cubernetes/ft-transcendence/pull/86/files#diff-d18e5a4d1e1ef6a3f1b598e53a2ee6ccff5cc6a09d31515d8e7bf72844ed9f44): I haven't fully decided what to do with this yet. I might still need it for later. TBD.

Yeah it's only commented out so there's no typescript complaining and git has it as well. We can just keep it where it is for access until you decide whenever.

> * **`ui` refactor:** Based on your previous code, I can support you with the refactoring. Might take me more time than you, so just lkm.

Sure, sure. I am still mostly just playing around with it, I am still not quite content with the UI structure. I think my main goal would be to abstract away reusable components. But UI isn't that big of deal tbh because it probably always be a little bit messy. And many different ways and styles would produce the same result so we can refactor as we go if needed.

I think I'll leave UI be before merging. I'm currently moving types and schemas around to improve type safety. And it feels like there are some bugs when running a game. I think I'll try to make local/remote game work with the setup panel before merging if that's cool. Should just take tonight if I'm lucky.

darrenkuro:
Another proposal: we should not use root (#app) as the router containers, some dom elements should be persisted in memory and toggle visibilities or reactively managed by states; such as header, footer, but most importantly - canvas, which is attached by renderer so lost canvas = need to recreate rendering engine. 

Other than in theory very expensive to keep reinstantiating and disposing, it is very difficult to decide how and when it should be generated (start-game via ws is obviously too late for remote game, but there are no other good flags to use that I can think of..) and synced between players of their renderer ready state (it should be in the range of ms but problematic in theory to start game without checking validity of rendering engine). 

If persisted, then everything can be properly awaited (no socket without renderer, for instance.) So the order of the operation can be better controlled?

I feel like there are better ways out there to do this, but yeah.. couldn't think of anything better currently.

OwnerOfJK:
> Some people really don't like it. I am still on the fence but for a very modular instantiated object like an engine of sort, I think it's probably fine. But kind of a clutch, still trying to think of better ways to manage them.

Okay, I am okay with leaving it the way you set it up. Just took me some time to understand how it works together.

> ..., it is very hard to see which lines belong to "one unit" since there's no indentation, ...

Ahh right, this makes sense to me. I had a similar feeling, but wasn't quite sure how to best optimise it. Feels a little like applying the tailwind method to entire html elements. So while a bit extra it feels good.

> I don't really know how blockchain stuff works but I think that makes sense. Feels like most of it are business layer logic and not much UI.

Yep, exactly. I will keep the ui part of it to a minimum. The main task here is to deal with the node provider.

> I think I'll try to make local/remote game work with the setup panel before merging if that's cool. Should just take tonight if I'm lucky.

Yes thats totally fine. Feel free to let me know if you would like me to assist.

> ... but most importantly - canvas, which is attached by renderer so lost canvas = need to recreate rendering engine. ...

Ah yeah I can issues where the user continously clicks between pages, reinitialising the game engine every single time. What you are saying is to handle it like you are already doing in `layout.store.ts` with the `header` and `footer`? Launching an initial state of the engine and renderer with default values, that are being adjusted with user input (game mode, game difficulty etc.) upon game launch.

In anticipation of a user launching a remote game, then launches a local game and then an AI, would we need one instance of the renderer (that handle the websocket communcation) per game mode? So we would initialise all of them at the start? This doesn't seem that optimal.

In general def agree that we should preload all resources we need for any game instance which includes sounds, textures etc.

darrenkuro:
> In anticipation of a user launching a remote game, then launches a local game and then an AI, would we need one instance of the renderer (that handle the websocket communcation) per game mode? So we would initialise all of them at the start? This doesn't seem that optimal.

I was kind of imagining that there could only be one game session at a time, so launching a game session will either stop the current game whether local or remote; or just don't allow another game session. Allowing multiple sessions is viable potentially but might be a little more complicated, but more importantly I am not sure if it makes sense, should a single user (one person) be allowed to have multiple sessions simulataneously?

Also, that check will be done at user level with auth state check I think. As for engine, my understanding is that it will be instance per tab. Since there won't be two canvases on one tab, it should be okay even if we allow it?

OwnerOfJK:
`I was kind of imagining that there could only be one game session at a time, so launching a game session will either stop the current game whether local or remote; or just don't allow another game session.`

Lets go with this. If the user attempts to leave/close the tab we should add a pop-up that essentially warns the user that if he leaves the page, the game will end (i.e. he/she will lose the game).

> Since there won't be two canvases on one tab, it should be okay even if we allow it?

Yep should be fine.

darrenkuro:
Okay, I think for the most part this PR is structurally sound as far as I am concerned.
There are still a large amount of work left to refactor especially for the UI stuff (I swear CSS is the hardest part in coding....). There are some UI stuff that's kind of broken visually currently since it's a WIP, but I don't think that matters too much and we can potentially fix and/or change them as we go. 

The persisted components are for the most part working on a foundational level, but could be buggy and still needs a lot of properly management of cleanup and moving around of state events. (I feel like there are still many events aren't being called in the exact right place) 

I am happy to keep working on it to get a much more polished and clean version; but let me know if anyone wants me to merge for any reason.

darrenkuro:
Sounds good. 

When implementing tournament UI, maybe try not to think of it as a page but rather as a modal, so just build a function that returns a container. That would be probably be the easiest way to integrate and restyle if needed, and it is modular so shouldn't interfere. And in that lane feel free to work on any new UI components, not sure if we need much more though, probably only user profile/dashboard to see stats and set up totp, upload avatar or something.

Blockchain should not interfere whatsoever so feel free to do whatever you want with that as well.

darrenkuro:
Mostly ready to merge now. Basically done refactoring setup page and modals. CSS is so difficult to test, a small bug can easily cost hours...

Anyways. No major update, only thing is I changed game to full screen for now because for w-full h-full fix either engine or css or something fixes the size ratio and often leave empty space vertically for some reason and I couldn’t figure out why. Also, that's probably better UI anyways? Should lock user into the game session until it finishes, or can have key hooks to exit if really desired. And it will get rid of window scroll interference issue for free.

OwnerOfJK:
Cool. 

> No major update, only thing is I changed game to full screen for now because for w-full h-full fix either engine or css or something fixes the size ratio and often leave empty space vertically for some reason and I couldn’t figure out why.

Had issues with that too. I think its fine. Don't worry about the visuals (i.e. CSS) for now. Its another rabbit hole that shouldnt be part of this refactor anyways.

I was just trying to run your branch and am getting 502 Bad Gateway error when I attempt to register. Not really up to date with the 2FA. Is there any obvious reason why this fails?

```
backend   | [09:01:48 UTC] DEBUG: Failed to create user
backend   |     error: {
backend   |       "name": "SqliteError",
backend   |       "message": "table users has no column named temporary_totp_secret",
backend   |       "stack":
backend   |           SqliteError: table users has no column named temporary_totp_secret
backend   |               at Database.prepare (/backend/node_modules/better-sqlite3/lib/methods/wrappers.js:5:21)
backend   |               at BetterSQLiteSession.prepareQuery (/backend/node_modules/src/better-sqlite3/session.ts:50:28)
backend   |               at BetterSQLiteSession.prepareOneTimeQuery (/backend/node_modules/src/sqlite-core/session.ts:135:15)
backend   |               at QueryPromise._prepare (/backend/node_modules/src/sqlite-core/query-builders/insert.ts:379:78)
backend   |               at QueryPromise.all (/backend/node_modules/src/sqlite-core/query-builders/insert.ts:396:15)
backend   |               at QueryPromise.execute (/backend/node_modules/src/sqlite-core/query-builders/insert.ts:408:40)
backend   |               at QueryPromise.then (/backend/node_modules/src/query-promise.ts:31:15)
backend   |     }
backend   | [09:01:48 UTC] INFO: request completed
backend   |     reqId: "req-1"
backend   |     res: {
backend   |       "statusCode": 500
backend   |     }
backend   |     responseTime: 299.4968330000411
```

darrenkuro:
> Cool. 
> 
> 
> 
> > No major update, only thing is I changed game to full screen for now because for w-full h-full fix either engine or css or something fixes the size ratio and often leave empty space vertically for some reason and I couldn’t figure out why.
> 
> 
> 
> Had issues with that too. I think its fine. Don't worry about the visuals (i.e. CSS) for now. Its another rabbit hole that shouldnt be part of this refactor anyways.
> 
> 
> 
> I was just trying to run your branch and am getting 502 Bad Gateway error when I attempt to register. Not really up to date with the 2FA. Is there any obvious reason why this fails?
> 
> 
> 
> ```
> 
> backend   | [09:01:48 UTC] DEBUG: Failed to create user
> 
> backend   |     error: {
> 
> backend   |       "name": "SqliteError",
> 
> backend   |       "message": "table users has no column named temporary_totp_secret",
> 
> backend   |       "stack":
> 
> backend   |           SqliteError: table users has no column named temporary_totp_secret
> 
> backend   |               at Database.prepare (/backend/node_modules/better-sqlite3/lib/methods/wrappers.js:5:21)
> 
> backend   |               at BetterSQLiteSession.prepareQuery (/backend/node_modules/src/better-sqlite3/session.ts:50:28)
> 
> backend   |               at BetterSQLiteSession.prepareOneTimeQuery (/backend/node_modules/src/sqlite-core/session.ts:135:15)
> 
> backend   |               at QueryPromise._prepare (/backend/node_modules/src/sqlite-core/query-builders/insert.ts:379:78)
> 
> backend   |               at QueryPromise.all (/backend/node_modules/src/sqlite-core/query-builders/insert.ts:396:15)
> 
> backend   |               at QueryPromise.execute (/backend/node_modules/src/sqlite-core/query-builders/insert.ts:408:40)
> 
> backend   |               at QueryPromise.then (/backend/node_modules/src/query-promise.ts:31:15)
> 
> backend   |     }
> 
> backend   | [09:01:48 UTC] INFO: request completed
> 
> backend   |     reqId: "req-1"
> 
> backend   |     res: {
> 
> backend   |       "statusCode": 500
> 
> backend   |     }
> 
> backend   |     responseTime: 299.4968330000411
> 
> ```

That's a dabatase issue from a while ago. So just do a make re to make sure it db is deleted and reinitialized properly. If it persists may have to manually inspect your docker volumes and clean everything.

OwnerOfJK (approved):
Checked and reviewed the code, ran the containers. Refactor looks good to me. Approved.

darrenkuro:
> Checked and reviewed the code, ran the containers. Refactor looks good to me. Approved.

I will do a final check and document progress in README then I'll merge this. 

@bejasch Maybe you can take a look after and we can discuss what parts that don't make sense to you that we should either roll something back or make other changes.

> 1. @darrenkuro continues on the refactor, and once satisfied, merges the PR.
> 2. @bejasch is working on the CLI component, which will be unaffected (however, it can definitely take inspiration from this refactor etc.).
> 3. @OwnerOfJK works on the tournament ui pages + integrating the blockchain component (while its inside the front-end repo, there shoudln't be any conflicts since its a new feature). I will pull form your branch as a starting point.

Also, if possible, let's stick to this plan at least until the next Meeting. 
There aren't a ton left and conflicts shouldn't be too crazy once this PR is merged but I might keep working on it a little.

PR #90, opened 2025-04-09
AI Player

yas0nia:
UPD: Agreed to merge this to resolve pong packages issues and then the comments will be addressed

Done:
- You can play against an "AI Algo"

Issues im resolving: even an easy mode is hard to play against xD 

Other issues I might resolve (not related to AI):
- When you return from a game after a few rounds and then choose another difficulty and click play, you return to the same score, which then returns to 0:0, but then after a score returns back to the old score...
- Winner is just "winner name"

P.S. I updated the logstash port to 5050 due to Mac issues Darren mentioned

### DIFF
--- a/frontend/src/scripts/modules/game/game.controller.ts
+++ b/frontend/src/scripts/modules/game/game.controller.ts
@@ -43,16 +43,16 @@ export const createGameController = (renderer: Engine, engine: PongEngine) => {
 
     const handleKeydownAi = (evt: KeyboardEvent) => {
         if (window.cfg.key.up.includes(evt.key)) {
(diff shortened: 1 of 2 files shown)


darrenkuro:
I think the AI service should be encapsulated entirely in the pong package as have the setting be passed in as config (perhaps as a boolean i.e. aiMode = true | false) and not expose to any additional logic to backend (or frontend for that matter), so that the client can configure it as needed wherever and whoever they are. That way you get AI for free in CLI for example, and it just makes sense to me regarding encapsulation.

darrenkuro (approved):

PR #93, opened 2025-04-15
Autoprune when deploy and allow http (8080) only on localhost

cubernetes:
Also, properly support multiple domains for dev mode (i.e. when you add them in /etc/hosts). Note that this doesn't work for .app domains, see comment in config.env.


darrenkuro (approved):
I think this looks fine, although I am not sure in what case you'll actually need multiple domains for dev.

cubernetes:
Yeah I was testing how well it works when you put it in /etc/hosts, mostly robustness testing I guess, encountered some interesting issues with HSTS, for example that you literally have to press the keys 'thisisunsafe' when you are on the page to bypass that setting, quite funny

PR #96, opened 2025-04-15
Refactor/normalize all quotes

cubernetes:
Normalize all quotes in comments and add script to find quotes that don't fit style. Also add pre-commit hook that flags single quoted text and groff style quoted text

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,8 +21,8 @@ include config.env
 include Makefile.clean
 include Makefile.aux
 
-# This target is needed for legacy docker compose versions where there is no `--watch' flag for `docker compose up'.
-# The `dev' target is preferred.
+# This target is needed for legacy docker compose versions where there is no `--watch` flag for `docker compose up`.
+# The "dev" target is preferred.
 .PHONY: dev-old-compose
 dev-old-compose: check-env
 	$(MAKE) clean-frontend-volume ensure-secret-files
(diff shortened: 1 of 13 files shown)


cubernetes:
Works on my machine. Set the pre-commit hook to non-enforcing for now, since I don't know how well this will work on other machines and I don't know how annoying it might when you're "coding in the flow". Might turn into a pre-push hook in the future.

darrenkuro:
Normalizing quotes feels like the job of the formatter, this might be a bit of an overkill haha..
Also it's only for quotes in comments right? What is script/quote_exceptions for? Quotes in code will be caught by linter and formatter and all three are allowed depending on situation anyways.

cubernetes:
Yes yes, it's only for comments. For quotes in code, the linter does everything. But you know, if everything is forcibly uniform, it's just a nice feeling haha

darrenkuro (approved):
Not sure about the scripts and hooks (their scope, extent, and how they work) but as long as they are not enforced I suppose this is fine.

cubernetes:
Yup not enforced, the script just finds all lines in all files (excluding some bloated folders) that contains comments with and quotes. You specify what kind of quote you're searching for, or multiple. It cannot find these kinds of comments:
```
/* stuff on this line can be found
anything in between here cannot be found
stuff here will also not be found because the regex is broken lol I just realized */
```

But the rest is the majority

PR #97, opened 2025-04-15
Misc/improvements

darrenkuro:
- Add health check for backend and let frontend depends on it.
- Specify every port in an environment variable, very unsure about ELK. Live reload port, backend port should be fully functionally. Haven't touched vault yet, not sure whether it's better to do a sed or something in Dockerfile to pass on the env or if .hcl support environmental variables natively like Caddyfile somehow.
- Rewrite route so to move away from hash routing and window href redirecting. Now with history API to have full control so that there's no more double load, etc.
- Scene in renderer will now take in config so that can be changed on frontend. 
- Tiny change in Pong, to let reset take in config as a parameter, but didn't publish yet to not potentially mess with the AI stuff @yas0nia is doing.


backend/Dockerfile
```diff
@@ -10,4 +12,8 @@ COPY . .
 
 ENV WATCH=0
 
+# Use a script to customize port?
+HEALTHCHECK --interval=5s --timeout=5s --retries=5 \
+  CMD curl --fail http://localhost:${BACKEND_PORT}/healthcheck || exit 1
```
cubernetes, line 17: Is ` || exit 1` superfluous or does it **have** to exit with exit code 1? Should the other `curl --fail` healthchecks (elk) be tweaked as well so it's uniform?

backend/Dockerfile
```diff
@@ -10,4 +12,8 @@ COPY . .
 
 ENV WATCH=0
 
+# Use a script to customize port?
```
cubernetes, line 15: Wdym with script here?

caddy/Caddyfile
```diff
@@ -28,9 +28,9 @@
 	}
 	handle_path /api/* {
 		import waf
-		reverse_proxy http://backend:3000
+		reverse_proxy backend:{$BACKEND_PORT}
```
cubernetes, line 31: Can BACKEND_PORT have a default value or rather not?
`{$BACKEND_PORT:3000}` like this
There is afaik no way to make it crash early if BACKEND_PORT is not specified

darrenkuro: I think the default value bubbles down, right? The default is set in compose so at this level, it should be guaranteed to have one?

cubernetes:
Regarding the .hcl files, [envsubst](https://hopeness.medium.com/master-the-linux-envsubst-command-a-comprehensive-guide-3cbab6ac6214) is the perfect tool for that. I used it for the Caddyfile before, until I figured out that Caddyfiles support environment variables natively. But for this, envsubst is definitely the right tool for the job. If you want, I can branch off and make a PR into this branch to implement that?

caddy/Caddyfile
cubernetes: Yes yes, our entrypoint is always the makefile. There are too many dependencies taken care of in the Makefile(s) that starting it bare would be too much of a hassle to maintain

darrenkuro:
Could you check whether vault is set up okay @cubernetes 
I opt to do it at build time, which might be less ideal than runtime with entry point, but I was a little afraid to do it inside because it looks complicated. This way it is guaranteed to be exactly the same as before. Maybe you can put it into entry point to avoid using build time args later

cubernetes:
Sure can do! Can you elaborate on the issue with build args tho? Aren't they fine in this specific case?

darrenkuro:
I think it's probably fine too. It might be a little less clean? Since it needs to be passed separately on compose file.
```
        build:
             context: "./vault/"
             args:
                 VAULT_API_PORT: ${VAULT_API_PORT:-8200}
```

And there is an extra modification step in Dockerfile when copying. 
```
# Generate config.hcl from template to expand environmental variables
 COPY ./config.hcl.template /vault/config/config.hcl.template
 RUN envsubst < /vault/config/config.hcl.template > /vault/config/config.hcl
```

But it's actually mostly fine imo. I just don't know if this is a standard way to do it. I feel like for some reason, most people prefer to defer all business logic until build time but maybe that's just a baseless preference.

cubernetes:
Fair enough, but I think I'm gonna keep it like this then, I try to do as much work as possible in the build stage. But all things in the entrypoint cannot be done in the build stage, but this particular issue is totally fine in the build stage. However, porting it into the entrypoint would be more or less trivial (afaik), something like this:
```diff
diff --git a/vault/entrypoint.sh b/vault/entrypoint.sh
index 8c4ac41..a8af3b4 100755
--- a/vault/entrypoint.sh
+++ b/vault/entrypoint.sh
@@ -85,7 +85,9 @@ debug "Creating vault storage path at directory '$vault_storage_path'"
 mkdir -p "$vault_storage_path"

 _start_vault_server () {
-       vault server -log-format=json -config="$vault_config"
+       vault server -log-format=json -config - <<-EOF
+       $(envsubst "\$VAULT_API_PORT" </vault/config/config.hcl.template)
+       EOF
        #vault server -log-format=json -dev
 }

```

And then add it in compose.yaml as well as an environment variable of course, instead of args.
But yeah, I think it's fine in the build stage

darrenkuro:
I think we can merge this. 

The port issue for ELK is beyond my understanding as for now and really couldn't get it work properly, sort of. The main problem is logstash seems to have files directly volume bind into the container that I am not sure if support env expansion. Having a Dockerfile would potentially solve that. But I am unsure about how logstash works as a whole so maybe we'll delay that til late. Will Open an issue about this when merging.

PR #98, opened 2025-04-15
Fix quote checker script to work on macOS

cubernetes:
Used IFS and sed to create a command line with all the exclusions. Drop in replacement for --exclude-from. Has its limits on linux, sine there's a maximum size the command line can have (https://www.in-ulm.de/~mascheck/various/argmax/), but good enough

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -10,15 +10,16 @@ die () {
 
 # subshell to not clobber IFS
 format_staged () (
+	PRETTIER_BIN="backend/node_modules/.bin/prettier"
+
 	[ "$#" -lt 1 ] && die "format_staged function expects at least 1 extension as an argument"
-	command -v npx 1>/dev/null 2>&1 || die "Please install npx to run pre-commit hooks"
-	npx prettier --version 1>/dev/null 2>&1 || die "Please install prettier through npx to run pre-commit hooks"
+	command -v $PRETTIER_BIN 1>/dev/null 2>&1 || die "Please install prettier to run pre-commit hooks"
 
 	IFS='|'
 	if git diff --name-only --cached --diff-filter=d | \
 		grep -E "\.($*)\$"                        | \
 		tr '\n' '\0'                              | \
-		xargs -r0 npx prettier --list-different 1>/dev/null; then
+		xargs -r0 $PRETTIER_BIN --list-different 1>/dev/null; then
 		all_good=1
 	else
 		all_good=0
@@ -27,7 +28,7 @@ format_staged () (
 	git diff --name-only --cached --diff-filter=d | \
 		grep -E "\.($*)\$"                        | \
 		tr '\n' '\0'                              | \
-		xargs -r0 npx prettier --write
+		xargs -r0 $PRETTIER_BIN --write
 
 	if [ ! "$all_good" = "1" ]; then
 		printf '\033[33m%s\033[m\n' "Warning: Not commiting since some files were formatted with prettier"
--- a/config.env
+++ b/config.env
@@ -1,7 +1,7 @@
 # Important: Never put quotes anywhere: https://github.com/docker/cli/issues/3630
 # Although, in compose it should actually work: https://docs.docker.com/reference/compose-file/services/#env_file-format
 
-# When running 'make dev', the variables HTTP_PORT, HTTPS_PORT, DOMAINS, ...
+# When running "make dev", the variables HTTP_PORT, HTTPS_PORT, DOMAINS, ...
 # will be set depending on these variables:
 DEV_HTTP_PORT=8080
 DEV_HTTPS_PORT=8443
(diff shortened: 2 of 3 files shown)


cubernetes:
Hm, thanks for flagging. Not sure why it would be slow, node_modules is the only giant thing. I'll investigate and make a commit for the -S -R thing

darrenkuro:
Okay, not that slow but feels slower than it should be. Maybe you could add a timer in script to benchmark how long the entire pre commit hook is running? It is definitely noticeably slower but idk if it's because something is wrong or just the script itself. Feels like about 10-15 seconds-ish maybe?

cubernetes:
Hm, what is indeed slow for me is prettier. I'll add two little timers.
```
find . | grep -v node_modules | grep -v \\.git | wc -l
```
Show's there's only roughly 300 files to search, not a lot for grep, usually

darrenkuro:
It does work. But yeah, just kind of slow. I counted roughly for 20s. Definitely feels slower than before..

cubernetes:
Damn that's too slow haha, I'll investigate

darrenkuro:
Hmm something is bizarre... this `time script/find_quotes roff_dquote roff_squote squote` takes virtually no time, however if I comment out check1 in precommit hook then git commit is relatively fast again...

cubernetes:
Wtf

cubernetes:
It might be because I invoked prettier 3 times in the format_staged function. A single invocation of prettier is pretty slow, even it it's not doing anything. I can imagine that doing it 3 times, can be quiet slow, especially because the last 2 times are bascially idential (there are there to check if prettier actually did anything, I'm not sure why I needed that maybe it can be removed).

Or is even a single invocation of prettier slow for you?

cubernetes:
Maybe you want to run this on your machine:
```console
ft-transcendence$ time git hook run pre-commit
pre-commit hook finished without error

real    0m0.911s
user    0m0.483s
sys     0m0.126s
```

And maybe a make clean before that?

darrenkuro:
it was consistently around 15-20s total for some reason.. much faster with after clean with not staged files, tried change a random file now it's 20s total again.
` time git hook run pre-commit
frontend/src/scripts/global/router.ts 41ms (unchanged)
pre-commit hook finished without error
git hook run pre-commit  1.76s user 0.48s system 11% cpu 20.054 total`

I don't get why it became much slower when nothing seemed to change for formatting. Before it was a noticeably pause but definitely not this much time, if I were to estimate probably 2 to 3s at most 5. 

Asked chatGPT to optimize format_staged blindly, it does actually make it a lot faster. But again, not sure what has changed...

cubernetes:
Oh, can you show what chatgpt gave?

cubernetes:
Hm, does it get slow again if you change `prettier` to `npx prettier`

cubernetes:
user time is the time cpu spent in userspace. sys time is the time cpu spent in kernel. real time is the sum plus anytime spent in sleep (like with the sleep command) or uninterruptile sleep (like waiting for a disk/network interface)

darrenkuro:
> Hm, does it get slow again if you change `prettier` to `npx prettier` and add back `npx prettier --version` ...

Slightly but still faster, it sort of varies.

But npx is quite slow. To ensure consistent version instead of using `npx prettier` we should probably just use `backend/node_modules/.bin/prettier` which actually does make it much faster.

cubernetes:
I see, I was wary using npx prettier from the beginning, now we see the problem it has. Can I just change it to backend/node_modules/.bin/prettier and use the script chatgpt gave and call it a day? Would that fix it?

darrenkuro:
```
format_staged () (
	[ "$#" -lt 1 ] && die "format_staged function expects at least 1 extension as an argument"
	
	PRETTIER_BIN="backend/node_modules/.bin/prettier"
	command -v $PRETTIER_BIN 1>/dev/null 2>&1 || die "Please install prettier to run pre-commit hooks"

	IFS='|'
	if git diff --name-only --cached --diff-filter=d | \
		grep -E "\.($*)\$"                        | \
		tr '\n' '\0'                              | \
		xargs -r0 $PRETTIER_BIN --list-different 1>/dev/null; then
		all_good=1
	else
		all_good=0
	fi

	git diff --name-only --cached --diff-filter=d | \
		grep -E "\.($*)\$"                        | \
		tr '\n' '\0'                              | \
		xargs -r0 $PRETTIER_BIN --write

	if [ ! "$all_good" = "1" ]; then
		printf '\033[33m%s\033[m\n' "Warning: Not commiting since some files were formatted with prettier"
		printf '\033[33m%s\033[m\n' "         Run 'git add -u' to stage those files as well"
		printf '\033[33m%s\033[m\n' "         Commit again to force the commit (i.e. ignore files changed by prettier)"
		return 1
	fi
)
```

Something like this is actually perfectly fine to keep the change minimal. Still so bizarre that npx suddenly became much slower for whatever reason... It might be my setting for npm or something 🤔

cubernetes:
Yeah odd... I'll make a commit

cubernetes:
(Might take a bit, my environment is broken because I tried to install [darling](https://docs.darlinghq.org/build-instructions.html), to test out the macOS side. Now my terminal isn't opening anymore. If it's not urgent I'll do it later)

darrenkuro:
I made one, I think it can be merged now so that the pre-commit hook on main would be working.

darrenkuro (approved):

cubernetes:
Awesome, thank you darren. Let's observe if issues arise

PR #100, opened 2025-04-16
Change backend healthcheck parameters

darrenkuro:
I think the deployment failure is probably due to backend exhaust retries before logstash start-up finish and gets marked healthy?

Add start period of 60s which seems to be the amount of time logstash will be marked unhealthy if I understood it correctly. And adjusted some other parameters which I think makes sense. Tested locally and worked fine, but server will have a different speed. Hopefully this works?

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -12,7 +12,7 @@ COPY . .
 
 ENV WATCH=0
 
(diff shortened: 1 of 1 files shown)


cubernetes:
Let's try 👀 💯💯💯

cubernetes (approved):

PR #102, opened 2025-04-18
Feature/tournament

OwnerOfJK:
### What has been done:
- **Tournament**: Added `tournamentMode()` to `SetupModal.ts`. It specifies a simple flow of creating a tournament by entering amount of players and player names. Does not yet launch a game instanc. **This is a still a WIP**.
- **Profile Page**: Added `PlayerStatsModal.ts` which basically does a fetch on `user/me` to pull all player stats. It generates a graph to display player stats. Also not yet fully finalised.
- **Quickplay**: Added a quickplay mode that essentially skips past registration and allows the player to play the `localgame` and `aigame` mode.
- **Cleanup**: Refactored `LoginForm.ts` to comply with current coding principles, removed `shaded` tailwind css property.

### What still needs to be done?
- **Tournament**: Full tournament flow to match registered players with each other and play games to determine winner.
- **Profile Page**:
   - Implement [`backend/route/me`](https://github.com/cubernetes/ft-transcendence/tree/backend/route/me) branch which provides API Route to player stats + player rank.
   - Add more styling to the page.
- **Quickplay**: Align with @darrenkuro for robustness.

I would like to merge this PR as it allows a full user flow end-to-end.


darrenkuro:
I think for tournament, the choice should probably be only 4 or 8, and maybe 16 (too many names to put in maybe?), So maybe as an input just be two or three buttons.

Because as discussed previously, knockout tournament probably is the mode that makes sense, round robin would take way too many matches when played one at a time. To avoid bizarre logic for seeding/matchmaking issue, none 2^n numbers are no good. And with 2 players it's really just one game and not a tournament.

darrenkuro:
And for online mode, is it better to implement a lobby system or keep it as is? Either way feels kind of fine, probably needs another layer of UI (with lobby, it'd be setting config and getting an room id; the way it is would be adding something to use when a user is waiting for an opponent.)

Maybe don't remove the onlineMode completely yet?

darrenkuro:
And unrelated to this PR, should quickplay also be entered via landing page (so that music plays)? 

Basically the correct setup with caddy guarantees user first enter landing page no matter the entry url. (even something like `/quickplay` that is unprotected) because it will already redirect to cfg.url.default (which is the landing page currently.)

It can be changed so that unprotected routes are actually accessible via url, but I'm not sure if that's desired behavior.

darrenkuro:
> Current solution is more of a placeholder. From a player perspective I would def prefer to be able to play against a chosen opponent (i.e. lobby with room id). If we would like to go with simplicity, I can add a page. What do you prefer?

Having a lobby system makes sense. 

> > And unrelated to this PR, should quickplay also be entered via landing page (so that music plays)?
> 
> Yes it should. I don't think any routes should be available via URL but can only be navigated to. The essentially means that the browser return button essentially makes an non-auth user jump back to `landing` right?

Okay yeah sounds good.

I think this is fine to be merged then whenever you wish to.

darrenkuro (approved):

PR #107, opened 2025-04-22
Increase start_period of logstash because it's slowwww

cubernetes:
Seems fair?

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -156,7 +156,7 @@ services:
             interval: 10s
             timeout: 5s
             retries: 3
-            start_period: 30s
+            start_period: 60s
 
     kibana:
         profiles: [elk]


darrenkuro:
Do we think we should change in logstash or the start period of backend? If it says logstash healthy but backend fail on the log, doesn't that suggest we should change it on the healthcheck on the backend

cubernetes:
But if backend is failing because of logstash, then we must give logstash more time to start up (in order to turn healthy), so that when the time comes, backend can start and encounter a healthy logstash service. After that, backend can do whatever trivial healthcheck that it wants to do (`curl --fail /healthcheck`, which will probably always success)

darrenkuro:
Does `Container logstash  Healthy` not mean that logstash has a successful healthcheck at some point once already?

cubernetes:
Hm, true that... I'm confused. Do we want to merge this and see how it works out, and then do another PR where we test the other way?

darrenkuro:
I'm afraid that won't prove or disprove anything, since #102 passed and #106 failed (which is kind of lucky?), I think we can safely say that's there's a racing condition somewhere in Dockercompose that makes it arbitrarily fail or pass. Probably difficult to completely be sure of this without digging deeper into how Docker/docker compose works :/

darrenkuro:
But just incase you want to see log or anything I'll prove for now

darrenkuro (approved):

cubernetes:
You're probably right, although my curiosity still wants to try both. In the 2nd PR, I'll revert this one and keep your proposed change

PR #109, opened 2025-04-22
Test/tweak healthcheck again

cubernetes:
Let's see if this behaves differently

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -12,7 +12,7 @@ COPY . .
 
 ENV WATCH=0
 
-HEALTHCHECK --interval=10s --timeout=2s --start-period=60s --start-interval=1s --retries=3 \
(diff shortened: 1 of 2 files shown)


darrenkuro (approved):
Btw just out of curiosity, there is no way to rerun a job manually without the trigger for GitHub action?

cubernetes:
There is, I just forgot to press it lol:
image

cubernetes:
But I'm currently logged onto the server trying to figure out why docker compose down doesn't work...

darrenkuro:
Honestly I have a suspicion that Docker is a very bug-prone piece of software...

cubernetes:
I started to notice that when you had those issues on Mac with the hot reload/watch. I've started to firmly believe that when I say that they don't do .env/dotenv parsing of variables correctly... And it only gets worse lol.

PR #110, opened 2025-04-22
Make down should also down elk

cubernetes:


darrenkuro (approved):
I am not sure if this actually fixes anything but I suppose at least it doesn't hurt?

cubernetes:
Yeah shouldn't fix the issue we were having, but doesn't hurt yes

PR #114, opened 2025-04-24
Feature/language

bejasch:
Added a languageStore in global/language.ts to subscribe to for a state change with the languageButton.

It doesn't rebuild all the elements but changes the .textContent.

The languageButton is still in the Header region, which is not clean, because it doesn't separate visually from the other navigation elements AND is not available on the landing page.

Optimisation TODO: Add the languageButton as standalone in some upper corner and have a flag symbol of the chosen language.


OwnerOfJK (approved):
Checked the code, and ran the container, all worked. Looks good to me.

Agreed that a flag ui would be nice to toggle between the languages. Could you maybe also add a explanation somewhere briefly explaining how to best add html components to the language store and what to pay attention to?

That way I can easily add new ui elements to it, and avoid a mess later.

PR #115, opened 2025-04-27
Feature/cookies

### LINKED ISSUE #88: Switch away from local storage
Vulnerable to XSS attack, session storage is lightly better, but best switch to httponly cookie?

Also, jwt-token expiration sort of randomly, what if a token expires mid-game or something?

Investigate more, maybe... refresh-token strategy?

darrenkuro:
- Fixes #88 , mainly. Switch from localstroage to using cookies.
- Makes TOTP modular, changes places for some services that belong to auth imo, cleaner flow with login. And corresponding frontend changes.
- Cleans up schemas.
- Moves authentication for socket to initial handshake rather than send jwtToken (since switched to cookies and no long have access to token, also it just makes more sense.)

Opening this PR since branch is made for switching to cookies, will implement changes on profile etc. in another PR soon.


darrenkuro:
Not sure what failed security check...

The idea is store password and username in memory in between the first login attempt and the second (with totp token). From what I can find it is a fairly common practice, to use the same end point `login` for the same purpose.

cubernetes:
It sees this as a secret:
image
I ignored this as a false positive, let's see if it flags it again

darrenkuro:
Ended up renaming it completely to bypass the check. Seems fine now.

OwnerOfJK (approved):
Mainly went through the changes in the `frontend` and `pong` and everything looks good to me.

PR #122, opened 2025-04-29
Feature/tournament final

### LINKED ISSUE #118: AIDifficulty duplicate when starting game with game controller.
in the code below we pass in the AIDifficulty as a new parameter instead of updating the PongConfig.

```
      const aiMode = (ctn: HTMLElement) => {
      ...
              controller.startGame("ai", undefined, { aiDifficulty: difficulty });
      ...
      }

    const startGame = (
        mode: GameMode,
        config: PongConfig = defaultGameConfig,
        options?: { aiDifficulty?: AIDifficulty }
    ) => {
        hidePageElements();
        hideRouter();
        showCanvas();

        switch (mode) {
            case "local":
                startLocalGame(config);
                break;
            case "online":
                startOnlineGame(config);
                break;
            case "ai":
                startAiGame(config, options?.aiDifficulty || "MEDIUM");
                break;
            default:
        }
    };


    const startAiGame = (config: PongConfig, aiDifficulty: AIDifficulty = "MEDIUM") => {
        engine.reset({ aiMode: true, aiDifficulty });
        attachAiControl();
        attachLocalEngineEvents();
        engine.start();
        startRenderer(config);
    };
```

### LINKED ISSUE #119: Button default onclick behaviour is overwritten by debug print
In `SetupModal.ts` the callback function is overwritten by a custom callback makes the button faulty.

```
const createDifficultyGroup = () => {
...

    const difficultyKeys = ["easy", "medium", "hard"] as TranslationKey[];
    const btnLabels = difficultyKeys.map((key) => getText(key)); // Get button labels as strings
    const btnCallbacks = difficultyKeys.map((key) => () => {
        window.log.debug(`Selected difficulty: ${key}`);
    });

    const btnGroup = createButtonGroup(btnLabels, btnCallbacks);
...
```

OwnerOfJK:
## What has been done:
This PR essentially consists of:
1. The first working MVP of the tournament system:
    - [tournament.controller](https://github.com/cubernetes/ft-transcendence/pull/122/files#diff-7051fb96bb169789cf7c4366c95c91dd26f82e1e0de56b2b913ef3363df028f7): Maintains tournament state and updates based on match results.
    - [contracts.interactions](https://github.com/cubernetes/ft-transcendence/pull/122/files#diff-af39507d430016cfa0fc6049c13e209c77ea3b4e0987e2847bfbb2642e3ac42d): Handles all blockchain interactions.
    - [TournamentPage](https://github.com/cubernetes/ft-transcendence/pull/122/files#diff-fd8ab1578bd1560b6bdb6144d8fb36425dfb4ecc3e419af3f676b3b8bcbec1f7): Renders the tournament page that includes all tournament HTML elements.
    - [TournamentBracket](https://github.com/cubernetes/ft-transcendence/pull/122/files#diff-283069685716211089730cb193130023ef2d1ad8644be57bc873e45a055d265d): Creates and updates the tournament bracket based on the tournament state.
    - Dynamic player names for "tournament", "ai" and "local" mode and dynamic winner return name.
3. Various front-end bug fixes.
   - Close #119 
   - Close #118 
4. A minor update to the pong-core package and game controller.
   - [pong.engine](https://github.com/cubernetes/ft-transcendence/pull/122/files#diff-f8d36652d67e6b655fd37ea88e28d73daf9b59c4304f82fe19fa01a34ad5e420): Include tournament mode and send the PongState at game-end.
   - [game.controller](https://github.com/cubernetes/ft-transcendence/pull/122/files#diff-bf5889a0958c9f32e198d02d31b106ba0eab22dd0d22f19e5ea1e2ac2f871ab4): Custom `handleEndRoundMatch` function that interacts with the `tournamentController` to update the state based on who won.

## Biggest open issues:
The biggest issue with the current flow is:
1. Once the game is over, I was unable to find a fix to automatically navigate back to the tournament page. Currently still requires clicking the back back button in the browser and clicking on the tournament mode in the `SetupModal.ts`. The tournament state is persistent unless the user does a full refresh or returns to the landing page. **Optimally the `handleEndRoundMatch` function ends the current game and navigates to "tournament".**


frontend/src/scripts/modules/game/game.controller.ts
```diff
@@ -200,6 +212,17 @@ export const createGameController = (renderer: Engine, engine: PongEngine) => {
         // Dispose stuff right away? Probably not..
     };
 
+    const handleEndRoundMatch = async (winner: string, state: PongState) => {
+        const { controller } = tournamentStore.get();
```
darrenkuro, line 216: this is in the controller scope, you don't need to get controller; can just be `await handleEndMatch(winner, state);`

darrenkuro:
Regarding issue 1: engine isn't emitting a name, that is the logic for the event handler, I think we can set them uniform to always take the entire event payload and figure it out on the frontend (engine will return whether left (0) or right (1) player won, names and everything should be accessible whether it's online or local)

frontend/src/scripts/modules/tournament/tournament.store.ts
```diff
@@ -1,19 +1,32 @@
 import { createStore } from "../../global/store";
+import { createTournamentController } from "./tournament.controller";
 
-type TournamentState = {
-    round: "Quarter" | "Semi" | "Final" | null;
-    matches: string[][] | null;
-    current_match: string[] | null;
-    players: string[] | null;
+export type Round = "Quarter" | "Semi" | "Final" | "Default" | "End";
```
darrenkuro, line 4: If we only accept 4/8 players, why would there ever be "Default" used for round? If it's for empty init state, it probably should be null instead

frontend/src/scripts/modules/tournament/tournament.controller.ts
```diff
@@ -0,0 +1,150 @@
...
+
+        tournamentStore.update(partialUpdate);
+        window.log.debug("Next Round created: ", tournamentStore.get());
+    };
+
+    const startTournament = () => {
+        tournamentStore.update({
+            tournamentId: generateUniqueTournamentId(),
+            activePlayers: allPlayers,
+            controller: controller,
```
darrenkuro, line 44: This is returning the wrong controller (look at the scope, it is by the end of this file, right before return) so think of this as equivalent to static class which is not intended I believe, I will add another comment to where the update should be located to be able to get that controller instance

frontend/src/scripts/ui/layout/SetupModal.ts
```diff
@@ -203,16 +208,18 @@ const setParticipants = (ctn: HTMLElement, playerAmount: number) => {
...
                 return showErr("player_names_required");
             }
+            players.add(player);
         }
-
         hideErr();
-        window.log.debug(`Tournament Start Data: ${playerInputs}`);
-        tournamentStart(playerInputs);
+        const controller = createTournamentController([...players]);
+        controller.startTournament();
```
darrenkuro, line 221: Here is where the controller is created, and where it needs to be updated into tournament store.

frontend/src/scripts/modules/game/game.controller.ts
OwnerOfJK: here I am extracting the tournament controller, not the game controller.

the handleEndMatch is from the tournament not the game. I will rename for clarify.

frontend/src/scripts/modules/tournament/tournament.utils.ts
```diff
@@ -1,18 +1,54 @@
...
+
+    if (!currentRoundMatches || currentRoundMatches.length === 0) {
+        return false;
+    }
+
+    // A round is completed if every match has a winner
+    return currentRoundMatches.every((match) => match.winner !== null);
+};
+
+export const generateUniqueTournamentId = (): number => {
```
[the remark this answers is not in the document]
OwnerOfJK: This was my initial implementation, however, if I understood correctly uuid returns a string, and I need a number for the blockchain interaction.

Storing a string as an ID in the contracts is more expensive.

frontend/src/scripts/modules/tournament/tournament.store.ts
OwnerOfJK: Sounds good. 👍

frontend/src/scripts/ui/layout/SetupModal.ts
OwnerOfJK: Thanks for catching this. Took me a little to understand why this is an issue because it didnt throw any errors, but to fix the issue the below would be the correct implementation, right?

`SetupModal.ts`
```
        const controller = createTournamentController([...players]);
        tournamentStore.update({ controller });
        controller.startTournament();
        navigateTo("tournament");
```

`tournament.controller.ts`
```
    const startTournament = () => {
        tournamentStore.update({
            tournamentId: generateUniqueTournamentId(),
            activePlayers: allPlayers,
        });
        createNextRound();
    };
```

OwnerOfJK:
> Regarding issue 1: engine isn't emitting a name, that is the logic for the event handler, I think we can set them uniform to always take the entire event payload and figure it out on the frontend (engine will return whether left (0) or right (1) player won, names and everything should be accessible whether it's online or local)

My initial intuition is to add the player names to the GameState:

`game.store.ts`
```
type GameState = {
    players: string[];
    isPlaying: boolean;
    isWaiting: boolean;
    gameId: string | null;
    opponentId: number | null;
    index: 0 | 1 | null;
    controller: ReturnType<typeof createGameController> | null;
};
```

For the local mode it would then be something like:

`SetupModal.ts`
```
    const playBtn = createCtaBtn("setup_play", () => {
        const player1 = p1.value.trim();
        const player2 = p2.value.trim();
        if (!player1 || !player2 || player1 === player2) {
            return showErr("player_names_required");
        }
        hideErr();
        const { controller } = gameStore.get();
        if (!controller) {
            return showErr("initialize_controller");
        }
        gameStore.update({ players: [player1, player2] });
        controller.startGame("local");
    });
```

Then in: `handleEndGame`:
```
engine.onEvent("game-end", (evt) => handleEndGame(evt.winner, evt.state));
const winner = players[evt.winner]; //either 1 or 0.
```

On the babylon game we can then also easily add the names to ensure that the correct players play the correct side (cannot enfore this on local our tournament).

Would this implementation be fine for online?

frontend/src/scripts/modules/tournament/tournament.utils.ts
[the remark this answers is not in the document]
darrenkuro: The problem here is that although probability is quite low, it is in theory possible for a collision (by billions folds compare to industry standard uuid or ulid etc) esp since this is client side. I guess we could either ping backend to generate a unique ID (if encryption is not an issue can even just be incrementing from 1), or to somehow map/hash uuid into a number (might need to be BigInt) or ulid (already a number but base32, so need BigInt potentially too)

frontend/src/scripts/ui/layout/SetupModal.ts
darrenkuro: Yep, architecturally correct, the rest is implementation inside the those subscriber functions.

frontend/src/scripts/modules/tournament/tournament.ui.ts
```diff
@@ -0,0 +1,86 @@
...
+        window.log.error("Failed to initialize public or wallet client");
+        return createEl("div", "text-red-500", { text: "Failed to initialize wallet client" });
+    }
+
+    let account: `0x${string}` | undefined;
+
+    const connectButton = createButton("Connect Wallet", "mx-auto", async () => {
+        account = await connectWallet(walletClient);
+        if (account) {
+            connectButton.style.display = "none";
```
darrenkuro, line 24: why are these not using tailwind?

darrenkuro:
> Would this implementation be fine for online?
> 


Yeah All you need is 
```
    registerHandler(
        "game-start",
        ({ gameId, opponentId, opponentName, index }) => {
            const { displayName } = authStore.get();
            if (!displayName) return;
            const players = index === 0 ? [displayName, opponentName] : [opponentName, displayName];
            gameStore.update({
                players: [displayName, opponentName],
                isPlaying: true,
                isWaiting: false,
                gameId,
                opponentId,
                index,
            });
        },
```
in ws.controller.ts, very clunky but I'll refactor stuff after.

frontend/src/scripts/modules/tournament/tournament.utils.ts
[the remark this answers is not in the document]
OwnerOfJK: okay let me open an issue for this and fix this later on another branch. #123

frontend/src/scripts/modules/tournament/tournament.ui.ts
OwnerOfJK: I tried a tailwind add class implementation but didnt work so I sticked with js. will see how to make it work later in another branch. #124

darrenkuro (approved):
Fine to merge when you want to imo

PR #127, opened 2025-05-05
Profile, Lobby, Locale, and refactor

### LINKED ISSUE #105: 2FA setup flow
I had only totp_secret in the beginning, but if you look at the documentation for speakeasy, indeed, they demonstrate the usually flow by having a temporary secret. I didn't implement the temporary secret at first, but it bit me later down the line, because think about it. Everytime you go to the 2FA setup page, it will fetch the /api/user/totpSetup endpoint. This endpoint will ALWAYS generate a new secret, necessarily. But you wouldn't want that just going to that page resets your secret right? That's why you need a temporary secret, that may change anytime, but at the same time persist in the database. As soon as you verify the temporary secret on the totpVerifyInitial endpoint, it will permanently save that secret. I think that's the procedure that is easiest to extend/debug/maintain, but surely there's a more compact of doing it by only saving totpSecret and a flag, but yeah that's about it.

_Originally posted by @cubernetes in https://github.com/cubernetes/ft-transcendence/issues/80#issuecomment-2781489897_
            
I was thinking that we have a user profile page (already WIP on the frontend I think) that will display some stats and have user-related setting such as change password, displayName and such. And in that page there will buttons to turn 2FA on/off, either toggle you'll have get a modal of QR code and code-enter input field to confirm (which is basically packing what's currently in the setup page excluding header and footer to a centralized container); if you had it on it'd fetch known secret if not it'll fetch new secret, so that totp setup will never be fetched when you already have that on. 

That way there'd be no reason to have a temporary secret right?

Anyways, I feel like that's often the 2FA flow that I encounter.

### LINKED ISSUE #120: Mouse cursor not a pointer on header
At least on feature/cookies, when hovering over the different routes in the header, the mouse cursor is not a pointer (should be `cursor-pointer` class in tailwind I suppose, https://tailwindcss.com/docs/cursor). For the language selection however, it seems correct.

### LINKED ISSUE #126: Clearing DOM container without properly dispatch event
Anytime there's a `.innerHTML = ""`, there needs to have a `dispatchEventDown(router, new Event("destroy"));` of sort to make sure there is no subscriber and listener leak.

```
    // Functionally dispatch the event bubbling down to all children elements
    const dispatchEventDown = (parent: HTMLElement, evt: Event) => {
        // Dispatch the event to the parent first
        parent.dispatchEvent(evt);

        // Dispatch the event to all child elements
        parent.querySelectorAll("*").forEach((child) => {
            child.dispatchEvent(evt);
        });
    };
```
This currently lives in router, should probably make it a utils function, or write a reactive mount/unmount similar to react to automatically manage these events.

When mount/unmount related events are set up correctly, each component should encapsulate and scope its own subscriber and events. For instance, for translation, it would be something like:
```
export const createBodyText = (text: TranslationKey, tw: string = ""): HTMLElement => {
    const baseTw = "text-xl text-black";
    const fullTw = `${baseTw} ${tw}`;
    const bodyEl = createEl("p", fullTw, { text });

    const unsubscribeLang = languageStore.subscribe(
        ({ lang }) => (bodyEl.textContent = texts[lang][text])
    );

    bodyEl.addEventListener("destroy", unsubscribeLang);

    return bodyEl;
};
```

### LINKED ISSUE #130: When disabling TOTP, the profile page doesn't update immediately (you have to hit refresh)

### LINKED ISSUE #131: Updating TOTP code gives INVALID_TOTP_TOKEN
```
backend   | [18:06:11 UTC] INFO: incoming request
backend   |     service: "backend"
backend   |     env: "development"
backend   |     reqId: "req-h"
backend   |     req: {
backend   |       "id": "req-h",
backend   |       "method": "POST",
backend   |       "url": "/totp/verify",
backend   |       "headers": {
backend   |         "user-agent": "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/135.0.0.0 Safari/537.36",
backend   |         "content-type": "application/json"
backend   |       },
backend   |       "remoteAddress": "10.42.42.1"
backend   |     }
backend   |     tags: []
backend   | [18:06:11 UTC] INFO: request completed
backend   |     service: "backend"
backend   |     env: "development"
backend   |     reqId: "req-h"
backend   |     res: {
backend   |       "statusCode": 401
backend   |     }
backend   |     responseTime: 0.9334749951958656
backend   |     tags: []
```

### LINKED ISSUE #134: 4P and 8P in Tournament is suspicisouly small

### LINKED ISSUE #141: Music button is green at the beginning, but music is not playing (have to press once to make it right)

darrenkuro:
Draft. Mostly done for what I planned to do but still needs clean up here and there. 

### Major implementations
- Lobby system, both backend module and frontend UI. It is mostly REST API, with create, join, update, leave end points. Host has the ability to update config (only using playTo as a proof of concept), and when they leave guest will be kicked also (lobby ownership doesn't transfer currently). Some channels with web socket to keep lobby information consistent between players. Seems mostly to be working, only functionally though. I suck at styling, needs more tweaking. 
Screenshot 2025-05-05 at 8 57 29 AM
- Profile page, mostly worked on backend for all related data end points. Most notably 2FA TOTP setup/update/disable/enable are now correct. Will implement password change and display name, update profile pic on next branch (I'll leave this branch and move to a new one be so that it can build for review). Still needs more styling, also for next branch. (Ignore the color, I just needed visualization when dealing with tailwind or css..) Stats and games as well. 
Screenshot 2025-05-05 at 8 58 00 AM

### Major changes and refactor
- `window` is true global on client side. It is unnecessary to have that prefix on frontend, except in some cases for clarity (i.e. `window.addEventListener`), for most other things the recommended style is not to use prefix so deleted in most cases. 
- HTML components are now standardized with parameters as an option object (i.e. `createButton = ({tw: "" ,text: ""})` etc.) Pros: easy to extend, input order doesn’t matter, and parameters are skippable. Using twMerge to override conflicting tailwind classes. And Each component should handle all its native event (for example translation.) 
- And for translation. Renamed as locale to comply with standard. Centralized translation process to one single subscriber, any time a text is passed into a HTML compoonent, it will look up if key exists, if it does, will save key in attribute with a key indicating which filed to translate, for instance, text, alt, or placeholder. All will be handled internally so user don't have to think about it. I think the I18N keys should probably be all upper case to indicate its translatable-ness to avoid accidental key-name/string collision. In addition, now it will save language preference for the website on local storage. And for new user it'll check `navigator.language` for locale preference. 
- Also using UIContainer and UIComponent to differentiate whether it's a container or a collection (array) of HTMLElements (might need better naming...). Otherwise it could get confusing and might nest unnecessarily deep with multiple containers of containers. And layout components can truly be easily used across different places, for instance, totp form when logging and when setting up. 
- Changed config on frontend to `CONST` and applied all caps style to differentiate.
- Result<Data, Error> is suboptimal structure, Error should really just be a string (with standard enum code instead of random messages). Changing in action. 
- Instead of doing `innerHTML = ""` now with `replaceChildren` to safely remove things in container and replace with new elements (it will dispatch destroy event to everything it destroys.) 
- Refactored some layout components (header, setup, totp, login form) to be cleaner, some styles may look (unintentionally) somewhat different. 

### Minor, less impactful changes
- Rebuilt database migration, after changing some columns migration started failing, so I re-generated the migration files. There are no real data so it should be fine. I am however unsure how the server is set up regarding this.
- Changed backend reply send to a decorator instead, with all errors being mapped from code (strings like "BAD_REQUEST", "USERNAME_TAKEN") directly so it's uniform data, much easier to incorporate translation for example. 
- Some clean up and refactor on pong-core package, mostly over socket stuff (due to changing to lobby system). Still ongoing. 

### Issues closed
- Close #126 
- Close #120 
- Close #105 
- Close #134 
- Close #130 
- Close #131 
- Close #141


darrenkuro:
Fix #134

darrenkuro:
I think this could be merged after review and I will keep working on more implementation on profile (avatar / change) on another branch.

Important design note: I believe all ui components should be uniform and therefore truly reusable (check totpForm for example, it's used for both injecting to a modal for setup/update/disable) and replace login form during login. 

The current way I employed is functions that start with `create` will return an array of elements to be injected into a container, so all tailwind classes will be in relation to its parent container; it can of course just return an array of a single item of a container if it's very important to style itself, but note that nested containers should have very good reason why they have to be nested. Otherwise the caller (this would be a higher layer; so page is highest which will be injecting to router container, and page will create a bunch of layout and components, so basically all containers should be styled here where they are called.) `create` functions should not have any side effects (or as minimal as possible), because they should basically mean "give me an array of page elements of this layout/page to be injected when invoked somewhere else".

Functions with `switch` prefix will mutate existing content of a container, for instance, replacing setup container to of a specific mode. Container could either be passed or gotten by id. 

Components should all be encapsulated to handle its own event and subscriptions to states. And when clearing content, should never use innerHTML = "" because that doesn't dispatch destroy event will is essential to not have leaks on store subscribers and potentially other things that wasn't removed properly. (replaceChildren is the wrapper to basically replace content of a container so we don't have to remember to manually invoke destroy event every time.) Having a function wrapper is a nice middle ground imo between the complexity of rewriting everything into customElements vs having to manually unmount on every occasion.

OwnerOfJK:
Huge refactor!! Only was able find some time to properly go through it today.

At first, a few comments:
1. Good stuff on the profile and lobby pages. A few more comments on this below.
2. Even though the following broke some CSS styling here and there, I think this is a nice standardization.
> so all tailwind classes will be in relation to its parent container
3. I didn't really dive too much into the changes for the language too much, maybe @bejasch best comments on them.

I separated out a few issues in separate comments below.

OwnerOfJK (approved):

PR #153, opened 2025-05-18
Integrate ELK fully into vault

### LINKED ISSUE #99: ELK env port compatibility
The port issue for ELK is beyond my understanding as for now and really couldn't get it work properly, sort of. The main problem is logstash seems to have files directly volume bind into the container that I am not sure if support env expansion. Having a Dockerfile would potentially solve that. But I am unsure about how logstash works as a whole so maybe we'll delay that til late. Will Open an issue about this when merging.

_Originally posted by @darrenkuro in https://github.com/cubernetes/ft-transcendence/issues/97#issuecomment-2808915986_

### LINKED ISSUE #143: Add env -u LOGSTASH_HOST to Makefile somewhere
To make dev and dev-old-compose work without ELK

### LINKED ISSUE #144: Manage ELK creds with vault

cubernetes:
Finally integrates ELK secrets fully into vault. This was surprisingly fun, but quite a difficult challenge!
For example:
- Ridiculous shell hack (`eval exec env -- - "$env_params" '"${0#-}"' "$(...)"`) to pass environment variable that are not valid identifiers (e.g. `bootstrap.memory_lock`, the dot is not valid for variables, but indeed valid for environment variables) (which is common practice for elastic services..., but for example `dash` (aka `sh`) purges such environment variables upon startup, not cool)
- Healthcheck migration: Healthcheck must a child process of the initial `entrypoint.sh` process tree, otherwise it won't have access to the environment variables extracted from vault. This also required a crazy `curl` feature that even ChatGPT didn't immediately know about. See elasticsearch `startup.sh` for example.

Also does some small things along the way, like
- curl `--no-progress-meter` instead of `--silent` where appropriate
- enable `git commit --squash=/--fixup=` to work with commit-msg hook
- enable compose bake for faster build times
- disable auto-restarting of ELK and backend containers, because it doesn't make sense when they get their config from vault
- fix names of volumes in `Makefile.clean`, they were all wrong and even some were missing lol
- support for JSON path-like references in env.json

Closes #144 
Closes #143 
Closes #99


cubernetes:
@darrenkuro can you confirm that this closes #99 as per your expectations?

cubernetes:
PRO TIP for review:
In one terminal run: `make dev-elk` (potentially `make clean` before)
In another terminal run: `make wait`

This will wait for all services to be healthy. You can also tweak the target to open the browser directory.

Another PRO TIP:
Run `make get-env SERVICE=elasticsearch` to get the password for the elasticsearch interface

cubernetes:
Ok, ready to review. Hope it works on your machines @darrenkuro @yas0nia.

Makefile
```diff
@@ -1,5 +1,5 @@
 # Change to docker-compose if you need
-DC := COMPOSE_BAKE=false docker compose
+DC := COMPOSE_BAKE=true docker compose
```
darrenkuro, line 2: What is COMPSE_BAKE? Seems new, couldn't find a lot of info..

Makefile
```diff
@@ -98,3 +101,12 @@ install:
...
 	npm --prefix=backend test
+
+# Usage: make get-env SERVICE=elasticsearch
+.PHONY: get-env
+get-env:
+	$(D) exec vault sh -c 'SERVICE=$(SERVICE); wget --header="X-Vault-Token: $$(cat /vault/secret/root_token)" --quiet --output-document=- "$$(cat /tmp/vault_addr)"/v1/secret/data/$$SERVICE | jq --raw-output --color-output .data.data'
+
+.PHONY: wait
+wait:
+	wait_for_service () { service=$$1; [ -n "$$service" ] || { echo "Expected 1 argument"; return 1; }; i=0; while [ ! "$$(2>/dev/null docker inspect -f "{{.State.Health.Status}}" "$$service")" = "healthy" ]; do printf "\015""Waiting_for_service_$${service}_to_start%$${i}s" "" | sed -u "s/ /./g;s/_/ /g"; i=$$((i + 1)); sleep 0.3; done; echo; } && start=$$(date +%s); wait_for_service vault && wait_for_service elasticsearch && wait_for_service logstash && wait_for_service kibana && wait_for_service backend && echo "Waiting 15 more seconds" && sleep 15 && echo "Frontend: http://localhost:8080" && echo "OpenAPI:  http://localhost:8080/api/docs" && echo "ELK       http://localhost:5601" ; now=$$(date +%s); min=$$(( (now - start) / 60)); sec=$$(( (now - start) % 60)); echo "Took $${min}m$${sec}s"
```
darrenkuro, line 112: Since we already have a script dir, maybe we can make this a file instead of writing it in one line?

Also, this just let you know it's ready and how long it took, right?

darrenkuro:
Could you maybe elaborate a little on the design of env.json vs config.env? I thought env.json was for generated stuff, but if it saves regular vars like port, why wouldn't we say move backend port in there (or pretty much everything)? Is it due to accessibility, like what container needs to access what variable and which way is easier? I'm not sure if I see a conceptual difference between the two and seems only to be about what's easier to parse in which context.

Just wondering, feel free to talk about it later if it's too much to type.

elk/elasticsearch/Dockerfile
```diff
@@ -0,0 +1,77 @@
...
+
+set -e # exit on any error
+set -u # treat failed expansion as error
+#set -vx # for debugging
+
+### Customization Point 2 ###
+service=elasticsearch
+
+vault_token=\$(cat "/run/secrets/\${service}_vault_token")
+vault_addr=http://vault:8200
```
darrenkuro, line 24: Hardcoded port that needs to be replaced? Same with the other two Dockerfiles

darrenkuro:
> @darrenkuro can you confirm that this closes #99 as per your expectations?

Hmm, I am not sure. I was thinking that there is a centralized variable that you can change that will control the ports of ELK services similar to the one in vault, backend, live reload, etc. I don't think the ones in env.json is changeable? Since it's hardcoded in compose, as well as config/script files in ELK. 

It might be an overkill and these ports might be super standard so we could scrape that if it's too much work. (In which case maybe all other internal services just use a hardcoded port as well for consistency?) 

The reason why I wanted make it changeable originally was due to A) consistency with other containers, B) possible conflict that would be block the service for painless startup (manually changing ports in weeds of config files might be too difficult).

darrenkuro (approved):
Everything seems to be working fine as far as I could tell though. Approved since none of above was anything major.

elk/elasticsearch/Dockerfile
cubernetes: Created an issue!

Makefile
cubernetes: It's an opt in feature of compose that enables for parallelized builds (faster). I always had it explicitly disabled to get rid of the advertisement message docker compose tends to print, and also because it broke things for me when I was creating a CTF, but now I just enabled it and it seems to do what it promises to do.

Makefile
cubernetes: Good idea, I'll do that. And yes, it tells you how long it took, but more importantly you can edit the script to automatically open the browser for you, meanwhile you can be busy with other things.

PR #163, opened 2025-05-19
Update/frontend bugs

### LINKED ISSUE #137: When game is done, it doesn't go to tournament overview page

### LINKED ISSUE #146: Ethereum is explicit any
On `types.d.ts`, `window.ethereum` is defined as any, which is bad ts practice. Not sure which library is registering that but perhaps it provided a type?

### LINKED ISSUE #159: Background music and game music play simultaneously
Easy to pause when game music starts, but say if user turn off bgm in game, background music outside the game should be muted too probably. Maybe have that preference saved together as bgmEnabled?

Not exactly sure about the desired behavior here.

### LINKED ISSUE #162: Change blockchain network from eth holesky to avalanche fuji
Currently the tournament data is sent to the deployed network on the holesky network because it was more acessible. Before finalising the project, we will have to change it to the avalanche fuji test network though.

### LINKED ISSUE #172: Round titles are not correctly shown in the tournament mode

### LINKED ISSUE #173: Already played tournament games can be started again

### LINKED ISSUE #174: Chart pop ups do not work

OwnerOfJK:
This PR fixes various front-end bugs and updates the UI where necessary:
- Closes #162 
- Closes #159 
- Closes #146 
- Closes #172 
- Closes #173 
- Closes #174 
- Closes #137 

The tournament page UI was revamped to look better.


frontend/src/scripts/modules/locale/locale.translation.ts
```diff
@@ -196,7 +202,66 @@ export const TEXT_MAP = {
         fr: "Nombre de joueurs", // TODO: Check if still correct
         es: "Número de jugadores", // TODO: Check if stil correct
     },
-
+    TOURNAMENT_GAME_COMPLETED: {
```
darrenkuro, line 205: I think for short text there is no need to specify where it is used, and the key should be as close as to its English version. So for instance, "completed" can just be with a "COMPLETED" key because chances are it might be reused in a different context and with the same meaning it would just be the same word in different languages anyway; and if in the future the need arises to change it then more detailed version/keys can be created. Same thing with "Pending", etc. And I think emojis should be probably be passed down as a variable to be interpolated in the caller since they will be identical across languages and they are not exactly translatable (one emoji wouldn't magically change to another in a different language) so that if it would ever need changed, you only do it in one place instead of changing the emoji in all the language texts. I can do that though, it's pretty trivial; just open an issue and resolve this comment.

frontend/src/scripts/modules/tournament/contracts/contracts.constants.ts
```diff
@@ -1,5 +1,7 @@
 // Contract address for the GameHistory contract deployed on Holesky testnet
-export const CONTRACT_ADDRESS = "0xfeefE87429c5dAC98009D663f62F05499bDAD2AD";
+export const FUJI_ADDRESS = "0x66532bC8c0A4996Ae1BeB5e9b68A55f3014B28a2";
```
darrenkuro, line 2: I think these can be moved to constants in global so they are in one place. Also I am not familiar with the development side of blockchain contracts, is it normal for wallet addresses to be so public with clients?

frontend/src/scripts/modules/tournament/tournament.store.ts
```diff
@@ -22,7 +22,7 @@ export type TournamentState = {
 
 export const defaultTournamentState: TournamentState = {
     tournamentId: 0,
-    round: null,
+    round: undefined,
```
darrenkuro, line 25: I think null is the given version of undefined; so for undefined, the standard way wouldn't to be declare it here in default, but rather make round optional in type, so 
```
export type TournamentState = {
    tournamentId: number;
    round?: Round;
    matches: MatchState[][] | null;
    current_match: MatchState | null;
    activePlayers: string[] | null;
    controller: TournamentController | null;
};
```
And it would be functionally identical, I think; so then type Round wouldn't include undefined.

frontend/src/scripts/modules/tournament/tournament.utils.ts
```diff
@@ -27,7 +27,7 @@ export const determineRound = (matches: MatchState[]): Round => {
     } else if (matchCount > 2) {
         return "Quarter";
     }
-    return null;
+    return undefined;
```
darrenkuro, line 30: I think it make more sense to use Result<whatever type you are expecting, E>, E being ideally ErrorCode but a string (serving as error message is fine.)

There are some other functions here and in contract (setupWallet was the one I saw) that have similar issues. (Basically instead of a log error -> return null/undefined, it would be just return a result and error type info would be included in the return so there is no need to log and caller has all the information to handle it however, so that the function is completely modular, functional, and self-contained.)

darrenkuro:
How was the tournament tested? Since it doesn't redirect when game ends and navigating away will go back to landing page and reset all states I couldn't get to anything beyond the initial tournament page..

OwnerOfJK:
> How was the tournament tested? Since it doesn't redirect when game ends and navigating away will go back to landing page and reset all states I couldn't get to anything beyond the initial tournament page..
> 
> Maybe also fix #137 if possible? I feel like do a navigateTo(tournament) should be fine at the end should basically suffice?

The way I tested it by essentially using the browser "go back" button, which brings me back from the game page to the setup panel. Once you click the `createTournament` button again, it should bring you right back to the tournament state.

The reason the `navigateTo(tournament)` doesnt work because the game canvas is practically launched under "http://localhost:8080/tournament" so is basically doesnt get past: ` if (location.pathname === path && !force) return log.warn(`Already in ${dest}`);`. If I comment this out it works.

```
export const navigateTo = (dest: Route, force: boolean = false) => {
    const path = `/${dest}`;

    // Alread in this route, do nothing unless it's the first route
    if (location.pathname === path && !force) return log.warn(`Already in ${dest}`);

    history.pushState({}, "", path);
    renderRoute(dest);
};
```

Not super sure what the best way here is. Should any game be inside its own "http://localhost:8080/{gameid}"?

frontend/src/scripts/modules/tournament/contracts/contracts.constants.ts
OwnerOfJK: Makes sense, I'll do that.

These are the contract addresses (not the wallet addresses), and yes it quite usual for them to be public. They are publicly indexed anyways.

darrenkuro:
> Not super sure what the best way here is. Should any game be inside its own "[http://localhost:8080/{gameid}](http://localhost:8080/%7Bgameid%7D)"?

Ah, I see, you can just call `navigateTo(tournament, true)` then, the second parameter is exactly for this purpose (first load also requires bypassing that check)

frontend/src/scripts/modules/locale/locale.translation.ts
OwnerOfJK: Okay, I updated the key names. Would be great if you could take on passing down the emoji. Opened an issue: #178

frontend/src/scripts/modules/tournament/tournament.utils.ts
OwnerOfJK: I see the idea, but honestly sometimes a little confused when it makes the most sense to return Result<> and when not. Would be cool to talk about this a little.

Will adjust accordingly and also flag a few other functions where I wasnt quite sure.

frontend/src/scripts/modules/tournament/tournament.store.ts
OwnerOfJK: Adjusted in: https://github.com/cubernetes/ft-transcendence/pull/163/commits/aecce0f0ee20801174930832ebca4fe393c461cc

frontend/src/scripts/modules/tournament/tournament.utils.ts
OwnerOfJK: I reworked some of my error handling, mainly removing any "throws", in some cases handling it more elegantly with ui showing it, in other simply returning. Should be alright for now.

I also show your createStatus component which might also be helpful, but want to focus on other things for now, that take higher prio.

darrenkuro:
Tried a 4P tournament but for some reason, after two matches on semi, final automatically has a winner stored already, is this a known error?

Otherwise I think this can be merged when you want to.

darrenkuro:
> > Tried a 4P tournament but for some reason, after two matches on semi, final automatically has a winner stored already, is this a known error?
> > Otherwise I think this can be merged when you want to.
> 
> This is because the engine is sending the "game-end" event twice, the tournament basically interprets this as:
> 
> Player1 won the semis and instantly won the Final, because a new round is automatically generated once the current is finished.
> 
> So should hopefully be solved once your PR is merged.

Ah okay, yeah then it should be solved after merging.

darrenkuro (approved):

PR #166, opened 2025-05-19
Update/engine + renderer and various issues

### LINKED ISSUE #74: Engine issues
In a right angle, if a paddle is hit, the ball gets caught by it:

https://github.com/user-attachments/assets/7a649b3f-e818-4f71-8465-bbc88ad64263

Somehow, sometimes, the tick rate seems to speed up in the engine. 
This happened when reloading (rejoining?) a game:

https://github.com/user-attachments/assets/e1e3dc4f-a056-4898-b067-5c3da6bd4835

### LINKED ISSUE #77: Compress state data sent over socket
Not that speed is or would be an issue but later on when structures are more finalized, maybe then only send essential data needed over the socket, and maybe compress them into binary and have coder/decoder.

Potentially.

### LINKED ISSUE #121: Add exit button to babylon game canvas
Add a button that shows up with the `showGameOver()` function to that is called by `handleEndGame()`.

It should redirect back to the setup page.

### LINKED ISSUE #138: maybe update game dimensions (currently too easy)

### LINKED ISSUE #139: Fix game delay into start game

### LINKED ISSUE #149: PlayerNames and GameMode persist beyond the game session.
If I start an AI game and afterwards start a tournament the tournament game for some reason is also an AI game. This implies that the game mode persists accross the previous game.

This sometimes also happens to the player names, when I want to create a game lobby, the displayed names come from the tournament not from the username. This implies that the gameStore still maintains the names of the tournament participants and overwrite the logged in players name.

### LINKED ISSUE #150: game-end is being called twice in a tournament game
For some reason the handleEndGame is being called twice in a tournament game, which leads to issues with updating the TournamentBracket.

### LINKED ISSUE #160: NPM package vulnerability
See if something can be done about this on our end
```
added 606 packages, and audited 607 packages in 6s

111 packages are looking for funding
  run `npm fund` for details

6 vulnerabilities (5 moderate, 1 high)

To address issues that do not require attention, run:
  npm audit fix

To address all issues (including breaking changes), run:
  npm audit fix --force

Run `npm audit` for details.
npm --prefix=cli install
```

### LINKED ISSUE #167: Handle dropping out of game by navigating away or dropping connection (e.g. closing browser)

### LINKED ISSUE #171: Remove upload volume with vclean

### LINKED ISSUE #175: Countdown before game start

### LINKED ISSUE #177: Translation of game elements

### LINKED ISSUE #178: Pass down emoji as a variable for translation.
[frontend/src/scripts/modules/locale/locale.translation.ts](https://github.com/cubernetes/ft-transcendence/pull/163/files/f6a2d728e1d8b16fbfa9d6fe7caa6b0b3e558d65#diff-2f6c4875a31fd22ff920f31785af2313c185bfd2d27b5e5b8f58af3382357f85)
```
export const TEXT_MAP = {
        fr: "Nombre de joueurs", // TODO: Check if still correct
        es: "Número de jugadores", // TODO: Check if stil correct
    },

    TOURNAMENT_GAME_COMPLETED: {
```

"""
I think for short text there is no need to specify where it is used, and the key should be as close as to its English version. So for instance, "completed" can just be with a "COMPLETED" key because chances are it might be reused in a different context and with the same meaning it would just be the same word in different languages anyway; and if in the future the need arises to change it then more detailed version/keys can be created. Same thing with "Pending", etc. And I think emojis should be probably be passed down as a variable to be interpolated in the caller since they will be identical across languages and they are not exactly translatable (one emoji wouldn't magically change to another in a different language) so that if it would ever need changed, you only do it in one place instead of changing the emoji in all the language texts. I can do that though, it's pretty trivial; just open an issue and resolve this comment.
"""

### LINKED ISSUE #179: Wins and losses are not correctly updated.
As per requirement for the Standard User Management we have to show the losses and wins. Thus, I added the below to the ProfilePanel:

```
    // Games played
    const gamesLabel = createParagraph({ text: GAMES_PLAYED, tw: LABEL_TW });
    const gamesEl = createParagraph({ text: String(user.totalGames) });

    // Wins
    const winsLabel = createParagraph({ text: WINS, tw: LABEL_TW });
    const winsEl = createParagraph({ text: String(user.wins) });

    // Losses
    const lossesLabel = createParagraph({ text: LOSSES, tw: LABEL_TW });
    const lossesEl = createParagraph({ text: String(user.losses) });
```

However, after playing a few rounds I get the below:

Image

Games played is correctly updated but not the wins and losses.

darrenkuro:
Issues
- Resolve #74
- Resolve #138 
- Resolve #139
- Resolve #150
- Resolve #177 : add translation for game elements, however they are not reponsive but it should be okay since there should be no way user change language preference during a game
- Resolve #171 
- Resolve #149 : need to keep an close eye and do more testing
- Resolve #178 
- Resolve #179
- Resolve #167 : ongoing, sort of fixed conceptually but need more testing
- Resolve #121
- Resolve #175
- Resolve #160 

Change notes
- Change paddle shape and remove emission light so it's easier to see where you should hit and where you shouldn't. 
- Change the backend console log during dev to something minimal and human readable. Did not change anything regarding production, or when LOGSTASH_HOST is set so functionally should have no difference except for during development and looking at your console. (Spent like 6 hours tracking down a bug and it didn't even end up being a backend one but rather a frontend one, oh well....)
- Automatically log user out (not removing cookies, just to landing page when logging in another tab/browser/device) because otherwise there needs to be way too much checking regarding which client could play etc. On backend, currently ensures there's only one socket available per user.

Due to the stupid bug that I couldn't find for an entire evening, didn't really process much especially the friends thing. Regarding stuff that I did manage to do, many, many things are still extremely messy and unorganized, to be styled, so to speak. But at least seems to be functional for the most part. 

But I am too lazy to test more throughly, likely to have many small bugs here and there though so if anyone has time please feel free to test, particularly about the game part and lobby/online, game difficulty (maybe still need to tweak parameters), and all issues listed above. Also bugs big or small, when/in what cases it crashes or behave wrong or whatever, reply or make an issue.


OwnerOfJK:
> (Spent like 6 hours tracking down a bug and it didn't even end up being a backend one but rather a frontend one, oh well....)

What bug did it end up being?

Runs really smooth, we finally have a smooth start and end for a game. Also tested the tournament and all seems to run smoothly. This looks more or less final.

darrenkuro:
> What bug did it end up being?

In landing page the CTA button does two asynchronous init calls at the same time (create the game controller and log user in which creates the socket), but the socket actually needs the game controller to register online game handlers. So I just nested the async calls but the whole initialization process just feels more and more iffy. 

Since I was changing socket on the backend, I felt like I broke something but after tirelessly logging info on backend still couldn't find why. Turns out it was probably a bug that has always been there just wasn't discovered since online game was never really tested before especially on a reconnect (it only occur when you have cookies saved already).

PR #180, opened 2025-05-21
Update/profile stats

OwnerOfJK:
This short PR implements a toggle in the profile page to view the Chart Section, Match History and Friends:
Closes: #132 
Closes: #133

Also improved the video quality of the video running in the background in the landing page.

### Charts
Screenshot from 2025-05-24 01-08-56

### Match History
Screenshot from 2025-05-24 01-09-02

### Friends (As of now this is just dummy data, needs to do a mapping for each friend).
Screenshot from 2025-05-24 01-09-07


frontend/src/scripts/ui/layout/GameStatsSection.ts
```diff
@@ -0,0 +1,176 @@
...
+    const rows = reversedGames.map((game) => {
+        const isPlayer1 = game.player1Username === playerName;
+        const opponent = isPlayer1 ? game.player2Username : game.player1Username;
+        const playerScore = isPlayer1 ? game.player1Score : game.player2Score;
+        const opponentScore = isPlayer1 ? game.player2Score : game.player1Score;
+        const result =
+            playerScore > opponentScore ? "Won" : playerScore < opponentScore ? "Lost" : "Draw";
+
+        return {
+            date: new Date(game.finishedAt).toLocaleString(undefined, {
```
darrenkuro, line 90: I would pass the locale in actually, instead of undefined, use `localeStore.get().locale`, they probably do the same thing but that would be explicit so it's clearer what the parameter is. Also, do we think changing language should change the locale to its corresponding one as well? I think that'd probably makes sense but I am not exactly sure.

frontend/src/scripts/ui/layout/GameStatsSection.ts
```diff
@@ -0,0 +1,176 @@
...
+    const reversedGames = [...games].reverse();
+
+    const headers = ["Date", "Opponent", "Result", "Score"];
+    const rows = reversedGames.map((game) => {
+        const isPlayer1 = game.player1Username === playerName;
+        const opponent = isPlayer1 ? game.player2Username : game.player1Username;
+        const playerScore = isPlayer1 ? game.player1Score : game.player2Score;
+        const opponentScore = isPlayer1 ? game.player2Score : game.player1Score;
+        const result =
+            playerScore > opponentScore ? "Won" : playerScore < opponentScore ? "Lost" : "Draw";
```
darrenkuro, line 87: There really shouldn't be a draw, if we want to be comprehensive, that probably should be an "Error" instead so at least we (devs) would know. 

Also these should be translation keys, but that can be changed together easily later cause there are other portions that have the same issue. Can just raise an issue and not fix it here.

darrenkuro (approved):

frontend/src/scripts/ui/layout/GameStatsSection.ts
OwnerOfJK: removed the possibility of a draw and there is a general issue to update all translations, will go through that at the end.

frontend/src/scripts/ui/layout/GameStatsSection.ts
OwnerOfJK: added, changing timestamps based on locale is a bit overkill in my opinion if thats what you meant.

PR #182, opened 2025-05-23
Feature/everything into env.json

### LINKED ISSUE #161: Don't hardcode vault API port
Hardcoded port that needs to be replaced? Same with the other two Dockerfiles

_Originally posted by @darrenkuro in https://github.com/cubernetes/ft-transcendence/pull/153#discussion_r2095281007_

### LINKED ISSUE #165: Adapt Makefile to extract compose.yaml environment variables from env.json

### LINKED ISSUE #185: Pass BACKEND_PORT ref to ".compose" in env.json
Actually yeah, I was kinda in the mindset to fully isolate everything just for principle, but yeah since BACKEND_PORT is not a credential anyways, it really doesn't matter. I'll put a reference in ".compose". Thanks for noticing haha.

_Originally posted by @cubernetes in https://github.com/cubernetes/ft-transcendence/pull/182#discussion_r2104400447_

### LINKED ISSUE #188: Unnecessary bind-mount on compose.yaml?
Was looking for info on #186 and incidentally saw that for elk there are many read only file bind-mount I think it's remnants of when we built elk images directly on compose? We should probably move them to their corresponding Dockerfile now right?

```
            - "./elk/elasticsearch/config/elasticsearch.yml:/usr/share/elasticsearch/config/elasticsearch.yml:ro"
            - "./elk/elasticsearch/config/generate-certs.sh:/usr/share/elasticsearch/config/generate-certs.sh:ro"
            - "./elk/elasticsearch/config/startup.sh:/usr/share/elasticsearch/config/startup.sh:ro"
            - "./elk/elasticsearch/config/ilm-policy.json:/usr/share/elasticsearch/config/ilm-policy.json:ro"
            - "./elk/elasticsearch/setup-single-node.sh:/usr/share/elasticsearch/setup-single-node.sh:ro"

            - "./elk/logstash/pipeline/:/usr/share/logstash/pipeline/:ro"
            - "./elk/logstash/config/logstash.yml:/usr/share/logstash/config/logstash.yml:ro"
            - "./elk/logstash/scripts/import-logstash-ca.sh:/usr/local/bin/import-logstash-ca.sh:ro"

            - "./elk/kibana/config/kibana.yml:/usr/share/kibana/config/kibana.yml:ro"
            - "./elk/kibana/dashboards:/usr/share/kibana/dashboards/:ro"
            - "./elk/kibana/setup-dashboards.sh:/usr/share/kibana/setup-dashboards.sh:ro"
 ```

@cubernetes can you confirm that is the case and there are no reason why they have to be bind-mount that I am missing completely? 
If it's indeed the case, feel free to move it if you'd like or assign me to do it.

### PR DESCRIPTION
**Pull request by cubernetes**
Closes #165 
Closes #161 
Closes #185 
Closes #188 

Wow this took me many hours...
Uses quite a few tricks to accomplish the desired result, but now no more `.env` or `config.env` messyness!

- every service is now managed by vault, except vault itself.
- env.json has 3 "special" services, namely `.Makefile`, `.compose`, and `.vault`
  - the leading `.` means vault will NOT try to create roles, policies, or tokens for this service
  - therefore, these can be conveniently used as an arbitrary KV store, isolated, but if you want, you can reference other fields using the {{ref:...}} syntax (see examples inside the file)
- Makefile has special helper functions to get environment vars from `env.json`, including a special function called `compose-env`
  - This special function gets all environment vars from the special `.compose` and `.vault` service and injects them alongside the variables gotten from the `.Makefile` service. The difference is that these 2 are passed as-is, while the stuff defined in the `.Makefile` service gets transformed to new environment vars (e.g. `PROD_DOMAINS` turns just into `DOMAINS`)
  - update README.md
  - env.json also support b64 (for base64 encoded random bytes) and env (for taking stuff straight from the environment)

### CODE REVIEW

File: backend/src/core/config/config.plugin.ts [hunk 1]
```diff
@@ -9,7 +9,9 @@ import { AppConfig } from "./config.types.ts";
 /** NODE_ENV should be used as process.env.NODE_ENV to ensure dead code is removed by esbuild */
 const configPlugin = async (app: FastifyInstance): Promise<void> => {
     // Validate config integrity with zod schema
+    console.log(`process.env=${JSON.stringify(process.env)}`);
```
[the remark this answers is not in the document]
cubernetes: Made a commit

caddy/Dockerfile
```diff
@@ -30,8 +30,54 @@ EXPOSE 8443
 
 VOLUME ["/frontend/.local/share/caddy/", "/frontend/dist/", "/frontend/uploads/"]
 
+# Environment wrapper stuff for vault
+RUN apk add --no-cache jq curl bash
+
+COPY --chmod=755 <<EOF /entrypoint.sh
```
darrenkuro, line 36: Why use heredoc? Wouldn't have entrypoint.sh file and just copy here be cleaner?

Makefile.aux
```diff
@@ -1,52 +1,50 @@
...
-		read c;               \
-		cp .env.example .env; \
-	}
+define compose-env
+$(shell ./vault/replace_json_templates.py <env.json | jq -r '.[".compose"] | to_entries[] | (.key + "='\''" + ( if .value | type == "string" then ( .value | gsub( "'\''"; "'\''\\'\'''\''")) else .value end) + "'\''")') \
+$(shell ./vault/replace_json_templates.py <env.json | jq -r '.[".vault"] | to_entries[] | (.key + "='\''" + ( if .value | type == "string" then ( .value | gsub( "'\''"; "'\''\\'\'''\''")) else .value end) + "'\''")')
+endef
 
 # Kind of ugly we have to do HTTP_DOMAINS and HTTPS_DOMAINS here, but inside Caddyfile it's kinda impossible
 # to do conditional logic (prod/dev) and text transformation (addprefix)
```
darrenkuro, line 15: We can do that in caddy's Dockerfile right? TBH this is also fine, only just a tad repetitive

backend/src/index.ts
```diff
@@ -31,6 +38,33 @@ const app = tryBuild.value;
 try {
     const { port, host } = app.config;
 
+    try {
+        const backendPort = process.env.BACKEND_PORT ?? "invalid"; // Provide invalid port so curl's status is 3, clearly indicating something when wrong
```
darrenkuro, line 42: I would just use port defined above directly, that is fully parsed and guaranteed to be valid (other wise this process would have exited already)

darrenkuro:
Nice, this is pretty cool.

I think the we still can't directly define ports for elk for it to work (but I am not sure if that's in the intended scope of this PR anyway so just documenting) due to the following hardcoded ports
- `http.port: 9200` in `elk/elasticsearch/config/elasticsearch.yml`
- `server.port: 5601` and `elasticsearch.hosts: ["https://elasticsearch:9200"]` in `elk/config/kibana.yml`
- `xpack.monitoring.elasticsearch.hosts: ["https://elasticsearch:9200"]` in `elk/logstash/config/logstash.yml`
- `hosts => ["https://elasticsearch:9200"]` in '/elk/logstash/pipeline/logstash.conf'
- `http://localhost:5601` in `wait target of Makefile`
- `http://kibana:5601/api/status"` and `http://kibana:5601/api/saved_objects/_import` in `/elk/kibana/setup-dashboard.sh`

backend/src/index.ts
darrenkuro: And regarding the subprocess here, does it close after detaching etc. Not sure if this is the best way/place to do it (but unsure since I don't 100% understand the mechanism and reasons yet)? My concern is that during dev this process doesn't only run once (due to watch), it could run upto hundreds of times, would the subprocesses close correctly?

caddy/Dockerfile
cubernetes: Yes true I remember. I'll extract all entrypoints to files. Will also make the "wait" makekfile target a script.

Makefile.aux
cubernetes: You mean to put logic like the one that chooses between PROD_* and DEV_* should go into the Caddy Dockerfile? Hm, true, that would be possible. But mind you that something like DOMAINS must also be available to the backend. I.e. I can't put everything into the Caddy dockerfile

backend/src/index.ts
cubernetes: Ah true, I missed that I just pasted some chatgpt code that would spawn a subprocess.

cubernetes: Regarding the other comment, yeah, I didn't think too deeply about it, I was mega tired haha. But you're right, I found this also to be rather ugly. It's unfortunately the ugly reality of a design decision I've made long long ago, namely that stuff defined in env.json is only available during runtime! This is the reason why for example the healthcheck cannot be implemented inside the Dockerfile as a HEALTHCHECK directive. Only the index.ts process has access to config stuff!! That's why you can only put the healthcheck implementation near it. The same problem is there for the ELK containers...

A possible solution for the backend only is the just write the BACKEND_PORT into a tmpfile and then write a Dockerfile healthcheck that reads that port from file. It works for the backend, since BACKEND_PORT is not a credential. However for ELK this doesn't work since sometimes you're working with credentials and you wouldn't want to put those in a file.


What do you think about rather putting the port into a file? I think that would avoid restart problems

cubernetes:
Oh you're right! Well, fortunately I leanred during the process that the .yml config files now support environment variables out of the box! So it's child's play to make those dynamic. Will do! Thanks for noticing

Makefile.aux
darrenkuro: Oh I was addressing the comments here
```
# Kind of ugly we have to do HTTP_DOMAINS and HTTPS_DOMAINS here, but inside Caddyfile it's kinda impossible
# to do conditional logic (prod/dev) and text transformation (addprefix)
```

backend/src/index.ts
darrenkuro: Maybe I am misunderstanding the setup somewhere, but in if you give the refs to `.compose` in env.json wouldn't we be able to just have whatever variables that are needed in compose file and from there it can easily pass it to build process?

Makefile.aux
cubernetes: Oh I see, the github mobile app doesn't really highlight the green that much I thought you were referring to the entire block.

Yeah you're right! I'll try to make it work :D

darrenkuro: Cool cool, it was not GitHub mobile btw, I added the comment on the line but I think since it wasn't actually changed by this PR it pulled the entire block to the nearest change haha

backend/src/index.ts
cubernetes: Actually yeah, I was kinda in the mindset to fully isolate everything just for principle, but yeah since BACKEND_PORT is not a credential anyways, it really doesn't matter. I'll put a reference in ".compose". Thanks for noticing haha.

darrenkuro:
By the way, have you heard of docker swarm? I was checking if there are easier and more standard ways to exchange data instead of using local files, like `"./.secrets/frontend_vault_token:/run/secrets/frontend_vault_token"`. And that came up which seems pretty cool. [Doc](https://docs.docker.com/engine/swarm/secrets/)

Not saying that we need to implement it especially if it's a complex switch, just something interesting.

cubernetes:
Yup I know about swarm, but swarm is actually more akin to kubernetes (-> overkill). For secrets, you're right they have facilities that make it easier. But the doc you linked also shows how you normally do docker secrets (without swarm), which is not what I want since I want to be able to have writeable files (and docker secrets are read-only).

Actually, I'm not 100% sure how docker swarm makes it easier, I intuitively thought it's just normal docker secrets but more at scale or smth. But yeah we're not gonna switch at this point lol, esp. because I don't have more than 1-2 hours to focus on transcendence today and tmrw, and we want to eval on Sunday.

darrenkuro:
Yeah I realized it is kind of like Kubernetes, and upon learning a little more about it, we definitely can't switch to that to work for both development and production. (Swam is a pure deployment system for production apparently, and can't support `develop` in compose.)

What about tmpfs mounts (just learned about this also)? Is there any particular reason why that wouldn't be a good fit here?

cubernetes:
Interesting, didn't know about that fact about swarm.

Regarding tmpfs mounts, is there a specific thing you can link to? Tmpfs mounts are quite a general concept, almost every linux system has it in /tmp and /run, and it's often just part of the RAM mapped as a file system as far as I know. What's the advantage you see having such a file system mounted?

darrenkuro:
> Regarding tmpfs mounts, is there a specific thing you can link to? Tmpfs mounts are quite a general concept, almost every linux system has it in /tmp and /run, and it's often just part of the RAM mapped as a file system as far as I know. What's the advantage you see having such a file system mounted?

https://docs.docker.com/engine/storage/tmpfs/

ChatGPT told me it's better lol, I honestly didn't know about this mount type until an hour ago so I honestly have no idea, that's why I asked. Again, no real reason to change anything though, I am curious whether there are some reasons why this mount doesn't suit our needs here and isn't actually a viable option.
```
Why tmpfs is better
	•	In-memory only: nothing ever touches disk
	•	Ephemeral: disappears on container stop
	•	Safer defaults for secret handling
	•	Works across multiple services if using named volume with tmpfs
```

cubernetes:
Ohhh yeah from a security perspective right! Yeah, in that case it might be better actually, if it's a no-brainer I might incorporate it, I'll make an issue

darrenkuro:
Actually you should probably hold off with that, it is a better option on Linux, I think, but apparently it might be a problem or buggy on MacOS with how docker manages tmpfs mount. Per ChatGPT:
```
But on macOS:
	•	Docker might silently fallback to using a normal local disk-backed volume inside the VM
	•	There’s no hard guarantee you’re getting true tmpfs behavior
```

Since you don't develop on Mac I don't think there are ways for you to test it, and it's a very trivial difference imo. Feel free to still open the issue and I'll take a look into it later and potentially test on my computer and see if it actually works or not.

cubernetes:
Already have the issue. I'll reassign it to you

cubernetes:
Ok, final version I guess. Run on my side, addressed all the issues, comments, nitpicks from copilot, and issues from the previous PR. Hope it works on your machine. If you approve, I'll merge.

Please double check that I didn't forget any comments/issues

script/wait_for_services
```diff
@@ -0,0 +1,35 @@
...
+wait_for_service elasticsearch
+wait_for_service logstash
+wait_for_service kibana
+wait_for_service backend
+
+echo "Waiting 15 more seconds because kibana healthcheck is currently not working quite right"
+sleep 15
+echo "Frontend: http://localhost:8080"
+echo "OpenAPI:  http://localhost:8080/api/docs"
+echo "ELK       http://localhost:5601" ;
```
darrenkuro, line 29: I think this is wrong still? 

Might be slightly difficult to get it here? Need to go through Makefile I think?
Since it's only text I guess alternatively we can just print to user something like the variable name and remind it's the one they've set, maybe. Up to you.

cubernetes:
Ahh interesting. Yes, you likely have to pre build it. If you know a cleaner way please tell lol, this felt very dirty when I wrote it. I don't want to build curl from source 3 times, which is why there's this builder image. I also don't know how to map build dependencies, note that we currently have COMPOSE_BAKE=true, so it's doing these parallel builds. Not sure how it's done professionally

script/wait_for_services
cubernetes: Correct it's host:container.

What do you mean with your last paragraph?

darrenkuro: Oh currently, the caddy container in compose.yaml, the ports are `"${HTTP_PORT:-8080}:8080"` and `"${HTTPS_PORT:-8443}:8443"`, which was what made me realize my misunderstanding. 

I was just wondering if they should be changed to `"${HTTP_PORT:-8080}:${HTTP_PORT:-8080}"` and `"${HTTPS_PORT:-8443}:${HTTPS_PORT:-8443}"` since it should affect anything, and all other containers have their ports in this format (i.e. as a pair)

The only other place this change would probably need is in caddyfile, currently we have
```
{
	skip_install_trust # caddy is running as non-root, therefore CA trust installation won't work (and isn't needed anyway inside the container)
	http_port 8080
	https_port 8443
	order coraza_waf first
	{$CADDY_EXTRA_GLOBAL_DIRECTIVES}
}
```

And changing to the following should suffice, I think?
```
{
	skip_install_trust # caddy is running as non-root, therefore CA trust installation won't work (and isn't needed anyway inside the container)
	http_port {$HTTP_PORT}
	https_port {$HTTPS_PORT}
	order coraza_waf first
	{$CADDY_EXTRA_GLOBAL_DIRECTIVES}
}
```

darrenkuro: I see, well. I guess it doesn't matter then. 

Maybe I'll change it the other way (the external-facing port/host is the one defined, the one in contain just use a constant), the only one that it might affect is LIVE_RELOAD_PORT, which only requires a couple lines of change. 
This way at least in compose.yaml the port style is consistent.

Now that I am thinking about it, this also is also there in production. But frontend is not a running container in that case, it builds and exits, and there is no socket open when it's in production; only port forwarding defined in compose. That is fine, right?

cubernetes: Sure sounds good!

For you last question, yup it doesn't matter

Makefile
```diff
@@ -40,16 +38,16 @@ dev-old-compose: check-env clean-frontend-volume ensure-secret-files
 	@$(call dev-env,watch --no-up)
 
 .PHONY: dev
-dev: check-env clean-frontend-volume ensure-secret-files
-	@unset -v LOGSTASH_HOST && $(call dev-env, \
+dev: clean-frontend-volume ensure-secret-files
+	@unset -v LOGSTASH_HOSTNAME && $(call dev-env, \
```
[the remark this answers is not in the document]
darrenkuro: Yeah true, just making sure I understand it right and I'll remember to remove this with that issue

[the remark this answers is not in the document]
cubernetes: I already removed it haha

darrenkuro (approved):

PR #194, opened 2025-05-23
Properly wait for dashboards to import

cubernetes:


env.json
```diff
@@ -64,7 +64,7 @@
     },
     "logstash": {
         "LS_JAVA_OPTS": "-Xmx256m -Xms256m",
-        "DEBUG": "true",
+        "DEBUG": "false",
```
darrenkuro, line 67: changing this to false doesn't actually work for me, I had to change `log.level: info` in `logstash/config/logstash.yml`. Does this work for you?

cubernetes: And actually I'm gonna remove the env var then

cubernetes: Ok I pushed it to the other branch, but whatever

cubernetes:
Yeah, it's really minor. I'm just mega confused why the grep command doesn't work.
`grep -m 1 means, exi

PR #225, opened 2025-05-25
Update/UI revamp

OwnerOfJK:
This PR is basically an entire overhaul of the UI. I decided to make it stand out and more unique.

Involved a lot of change in styling accross all pages to fit into the new ui and also involves some changes in the UX flow.

Will provide more details.


frontend/src/scripts/ui/pages/LandingPage.ts
```diff
@@ -27,7 +26,7 @@ export const createLandingPage: PageRenderer = async (): Promise<HTMLElement[]>
 
     // Create a call-to-action button
     const ctaButtonEl = createButton({
-        text: CONST.TEXT.PLAY,
+        text: "Play",
```
darrenkuro, line 29: Language button is still available at landing page (as it should), I'm not sure why this is changed to string literal, simply switch back to CONST.TEXT.PLAY would restore language feature for this button.

OwnerOfJK: I did this on purpose to disable translations for the landing page. "Play" in other languages escape the ctaButton, I can see no reason why we would want to translate play. Its the action word to lead the user into a fully translated app.

darrenkuro: I see, well in that case I would just remove the language button from landing page instead

OwnerOfJK: Yep, agreed. Was thinking I had already done that, thanks for mentioning it.

darrenkuro:
Could you take a screenshot of how it is breaking? I couldn't really think of a reason why (except for the z index which is easily changed). It is of course not a huge deal, it's just feels too close to the form and personally more importantly it is on the different end than after you log in, maybe we move the user status container to the left corner?

OwnerOfJK:
Its super weird, from what I can see (the code) it should practically work but for some reason it gives me this (this is from the code above):

image

Checking why. And before that the result what this:

image

darrenkuro (approved):

PR #237, opened 2025-05-25
Fix/i18n element

### LINKED ISSUE #223: Remove unnecessary i18n setup when creating elements
Hunt down everything that looks like this, now createEl manages it internally so they are completely unnecessary 
```
        const link = createEl("a", "hover:underline cursor-pointer", {
            text: getText(key),
            events: { click: () => navigateTo(route) },
            attributes: { [CONST.ATTR.I18N_TEXT]: key },
        });
```

darrenkuro:
Fix #223 

Cleared some todos in place, raised some issues for others.


cubernetes:
Is it now fix/various-issues?

darrenkuro:
Yeah changing the branch name to reflect what it is doing, closed because idk how to like delete a PR completely

PR #242, opened 2025-05-26
Extract healthchecks to separate files

### LINKED ISSUE #215: Clarify method to do healthchecks
> And in this process I also realized something else, for the following block or similar ones:
> 
> ```bash
> # Healthcheck background service (because only this process has access to the environment
> # and an external healthcheck would need access to those, which is not intended)
> TERM=linux setsid -f watch -xtcn5 curl \
> 		--no-progress-meter \
> 		--insecure \
> 		--user "${ELASTIC_USER}:${ELASTIC_PASSWORD}" \
> 		--fail \
> 		--write-out '%output{/tmp/healthcheck}%{exitcode}' \
> 		"https://localhost:${ELASTICSEARCH_PORT}/_cluster/health" 1>/dev/null 2>&1
> ```
> This (comment) is no longer true, right, we can route through .compose in env.json to do all healthcheck the traditional way? However, here it uses user and password which I am not sure if it's a good idea to export those to compose (but using .env would do the same so shouldn't be a problem?)

(originally posted by @darrenkuro in https://github.com/cubernetes/ft-transcendence/pull/193#issuecomment-2907846016)

cubernetes:
Closes #215


cubernetes:
LOL, there's still a git conflict resolution marker

cubernetes:
Fixed haha

cubernetes:
Rebased as well

cubernetes:
Thanks!

PR #262, opened 2025-05-29
Safeguard player not sending ready timeout

### LINKED ISSUE #236: Handle timeout for when the other player sends in ready
Unlikely to happen since it would be handled separately if a player quit, but I am not sure whether there are some ways a player can be "stuck" or hacked and just never send it ready, so should also set a timeout range before declare whoever is ready the winner

darrenkuro:
Fix #236 

Logic should work in theory but not sure how to test this case in particular.


bejasch (approved):
Works fine.
Maybe add a var for the timeout time.

PR #268, opened 2025-05-29
Update dependabot.yml

cubernetes:


darrenkuro (approved):
Let's see what happens

PR #269, opened 2025-05-29
Make backend healthcheck more lax

cubernetes:


darrenkuro (approved):
Let's see if this happens

